PR #6, opened 2018-09-06
Enables Kotlin support with Gradle plugin for compilation and runtime.

ivanporty:
Enables compiling, testing and running Kotlin files possible with Gradle.


etanshaul:
check out: https://www.jetbrains.org/intellij/sdk/docs/tutorials/kotlin.html . specifically:

> Please note that you should not include kotlin-runtime and kotlin-stdlib jars with your plugin because Kotlin guarantees backward- and forward- binary compatibility

Is this applicable?

ivanporty:
You're right, they are not necessary, IJ was just acting up like it cannot run the test file, it needed some internal library setup, but the script works fine in the simpler form.

build.gradle.kts
```diff
@@ -3,12 +3,7 @@ repositories {
...
+    id("org.jetbrains.intellij") version "0.3.7"
 
     kotlin("jvm") version "1.2.61"
-}
-
-dependencies {
-    compile(kotlin("stdlib"))
-    compile(kotlin("runtime"))
-}
+}
```
etanshaul: nit: linebreak

build.gradle.kts
```diff
@@ -1,3 +1,9 @@
...
-    id("org.jetbrains.intellij") version "0.3.6"
+repositories {
+    jcenter()
 }
+
+plugins {
+    id("org.jetbrains.intellij") version "0.3.7"
+
+    kotlin("jvm") version "1.2.61"
+}
```
etanshaul: nit: line break

build.gradle.kts
```diff
@@ -1,3 +1,9 @@
-plugins {
-    id("org.jetbrains.intellij") version "0.3.6"
+repositories {
+    jcenter()
```
etanshaul: jcenter is a superset of mavenCentral right? (I remember reading that its more performant..so lgtm)

ivanporty: I believe so, yes, many claims it behaves more performing and reliable. We can always change it if we see something wrong with it.

PR #23, opened 2018-09-20
Add ktlint Kotlin linter via Spotless plugin

### LINKED ISSUE #16: Add Kotlin static analysis / code formatting tools

etanshaul:
fixes #16 

Spotless plugin (used in cloud tools for intellij also) has built in support for ktlint.

There are a few peculiarities that we can iron out as we go along:

- the IJ formatter doesn't seem to enforce the line length rule (even though the linter will fail since 100 char limit is part of the Android Studio style guide)
- The current settings can lead to some strange formatting (see the changes to `SkaffoldUiResources` in this PR). As far as I can tell this is correct Kotlin styling 
- ktlint has some basic auto-formatting limitations: it cannot (yet?) auto-format indentation problems, convert wildcard imports etc.. The IntelliJ formatter / import optimizer can fix these though.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
+    <option name="RIGHT_MARGIN" value="100" />
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value>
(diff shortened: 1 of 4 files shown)


etanshaul:
see the discussion here: https://github.com/yole/kotlin-style-guide/issues/15#issuecomment-225418198 w/regards to the formatting changes applied to `SkaffoldUiResources`.  I agree with yole's comment, but looks like they went with it:

https://kotlinlang.org/docs/reference/coding-conventions.html#function-formatting

etanshaul:
Here's the same rule in the Android style guide (which is the one we are following currently):
https://android.github.io/kotlin-guides/style.html#functions

PR #39, opened 2018-09-27
 Added search of Skaffold files across the project space.

### LINKED ISSUE #37: Add Skaffold file search across the project
Iterate project files, without excluded directories such as module compile output, and use the list of found files to populate drop-down with available Skaffold files in run configurations UI settings.

Part of #7, #12, #13.

ivanporty:
Fixes #37.

Implements recursive search of all Skaffold configuration files across the base project directory, skipping the excluded directories (i.e. build outputs etc.)

Unit tests are coming in the follow up PR - it will also need some initial test infrastructure to test virtual files and project.


ivanporty:
@tiembo quick glance at Kotlin in general here is appreciated!

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Finds all Skaffold configuration YAML files in the given project. "Excluded" directories from
+ * project modules are omitted and not searched (such as build output).
+ */
+internal fun findSkaffoldFiles(project: Project): List<VirtualFile> {
+    val results: MutableList<VirtualFile> = ArrayList()
+    val excludedRoots: MutableList<VirtualFile> = ArrayList()
+    ModuleManager.getInstance(project)
+        .modules.forEach { excludedRoots.addAll(it.rootManager.excludeRoots) }
```
ivanporty: @tiembo I was not sure how to collect this in Java `collectors` style, maybe you have a suggestion here?

build.gradle.kts
```diff
@@ -48,6 +50,12 @@ allprojects {
             ktlint()
         }
     }
+
+    dependencies {
+        testCompile("com.google.truth:truth:+") {
```
etanshaul: is this needed for this PR? if not, maybe include it in the next one when you add testing infra

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -0,0 +1,82 @@
...
+                // consider this YAML file as Skaffold when first line contains proper API version
+                return scanner.hasNextLine() && scanner.nextLine().startsWith(SKAFFOLD_API_HEADER)
+            }
+        }
+    }
+    return false
+}
+
+/**
+ * Finds all Skaffold configuration YAML files in the given project. "Excluded" directories from
```
etanshaul: I'm not sure what it looks like for kotlin, but maybe expand the doc to the full style (like javadoc, with params, return, etc.)?

etanshaul:
Also just a general question:

I think you hinted that the idea here is that the Skaffold run config would show a dropdown listing all of the available skaffold.yaml files.

At the same time, we plan to also have detection of `skaffold.yaml` files and autocreate the run configurations -> presumably it would then default the dropdown to the given `skaffold.yaml`. So I guess the idea is that a single run configuration is not tied to a specific skaffold.yaml and that a user can change it on the fly (instead of creating a new run config tied to a different skaffold.yaml).

I think this makes sense, just making sure I understand the approach.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -0,0 +1,82 @@
...
+import java.io.ByteArrayInputStream
+import java.io.InputStream
+import java.util.Scanner
+
+private const val SKAFFOLD_API_HEADER = "apiVersion: skaffold/"
+
+/**
+ * Checks if a given file is a valid Skaffold configuration file based on type and API version.
+ */
+fun isSkaffoldFile(file: VirtualFile): Boolean {
```
etanshaul: are you leaving this public on purpose? or is this to get around the `internal` issue with gradle test source sets living in a different module?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -0,0 +1,82 @@
...
+) {
+    if (file in excludedRoots) return
+
+    if (isSkaffoldFile(file)) {
+        results.add(file)
+        return
+    }
+
+    if (file.isDirectory) {
+        file.children.forEach { findSkaffoldFiles(it, excludedRoots, results) }
```
etanshaul: Instead of manually recursing through the file tree, could you use the IJ file index?

see https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/app-engine/java/src/com/google/cloud/tools/intellij/appengine/java/project/DefaultAppEngineAssetProvider.java#L64

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
etanshaul: maybe a question for @tiembo, but what about making this an extension function on `VirtualFile`?

From my reading it seems that Kotlin favors using top level extension functions in place of the Java utility style classes.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
patflynn: +1.  This doesn't seem to be the optimal way to find files in an IDE.

build.gradle.kts
ivanporty: agree, that's for the next one.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
ivanporty: I'll take a look at that, but the file name maybe in an arbitrary form. I'll check if it has an ability to check for a specific content and yaml file.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
ivanporty: Correct. let me bring back `internal` one and then we'll discuss this in the next PR.

ivanporty:
@etanshaul on your general question, correct, it feels run configurations naturally allow changing their Skaffold configuration file. Auto-creating run configs for Skaffold files will help with that, but it doesn't make sense to "hard code" one Skaffold file per run configuration.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -0,0 +1,82 @@
...
+    return results
+}
+
+/**
+ * Recursively finds all Skaffold configuration YAML files in the given root file/directory.
+ * @param file file to start search from, directory or a file itself.
+ * @param excludedRoots file roots to exclude search on.
+ * @param results list to collect resulting Skaffold files into.
+ */
+internal fun findSkaffoldFiles(
```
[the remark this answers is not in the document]
etanshaul: oh oops I didn't realize that. hopefully we can refactor this to 1) get rid of the recursion and 2) use more IDE internals for finding the files we need.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
etanshaul: sg

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
ivanporty: More on that: Skaffold file contents may not be indexed by IDE at all and not be a part of source file sets. Something to think about while this approach will find any Skaffold files in the project base directory. Checking file indexing APIs.

etanshaul:
cool. that makes sense to me. some users may prefer having a single run configuration and jsut changing its context rather than having many.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
ivanporty: added.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
etanshaul: one thing to check:

I recall that the framework support infrastructure which we used for our app engine stuff and is based on the presence of a file (`appengine-web.xml`) actually looks into the contents of the file and not just the file name. You can check to see what the logic is doing there.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -0,0 +1,82 @@
...
+ * Checks if a given file is a valid Skaffold configuration file based on type and API version.
+ */
+fun isSkaffoldFile(file: VirtualFile): Boolean {
+    with(file) {
+        if (!isDirectory && fileType is YAMLFileType && isValid) {
+            val inputStream: InputStream = ByteArrayInputStream(contentsToByteArray())
+            inputStream.use {
+                val scanner = Scanner(it)
+                // consider this YAML file as Skaffold when first line contains proper API version
+                return scanner.hasNextLine() && scanner.nextLine().startsWith(SKAFFOLD_API_HEADER)
```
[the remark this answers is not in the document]
ivanporty: Changed to the regex that should cover deviations.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.psi.search.FileTypeIndex
+import com.intellij.psi.search.GlobalSearchScope
+import org.jetbrains.yaml.YAMLFileType
+import java.io.ByteArrayInputStream
+import java.io.InputStream
+import java.util.Scanner
+import java.util.regex.Pattern
+
+// see https://github.com/GoogleContainerTools/skaffold/blob/master/examples/annotated-skaffold.yaml
```
etanshaul: nit: should this have a block comment style (`/* */`). I'm not certain myself

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
ivanporty: Changed to use File type based index so we won't depend on file name. For flexible detection, we use similar approach, actually reading in file content for each matching file, but with framework detection.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -0,0 +1,62 @@
...
+                    SKAFFOLD_API_HEADER_PATTERN.matcher(scanner.nextLine()).find()
+            }
+        }
+    }
+    return false
+}
+
+/**
+ * Finds all Skaffold configuration YAML files in the given project. "Excluded" directories from
+ * project modules are omitted and not searched (such as build output).
```
etanshaul: nit: should there be a line break between the comment and the param section? in javadocs there is

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -0,0 +1,62 @@
...
+ * @param project IDE project to search Skaffold file in
+ * @return List of Skaffold configuration files in the project.
+ */
+internal fun findSkaffoldFiles(project: Project): List<VirtualFile> {
+    return FileTypeIndex.getFiles(YAMLFileType.YML, GlobalSearchScope.allScope(project))
```
etanshaul: very nice refactoring

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
ivanporty: cannot see anything on this there, but clarified a bit and removed "excluded" mention.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
ivanporty: Didn't find anything in style guide. Maybe @tiembo knows some doc suggestions?

ivanporty:
Kotlin reviewers are at KotlinConf, and this is a small PR, so it looks as good to go.

etanshaul:
Feel free to merge, but we can revisit the question I had about extension functions and refactor later if needed

Issue #13, opened 2018-09-17
Create Skaffold run configuration and UI for continuous deployment / development.


ivanporty:
Run configuration allows a user to run Skaffold continuously (development mode, `skaffold dev`), and stream the logs into the console window until stopped by a user.

Includes:

- declaration of run configuration
- factory to create configuration
- settings
- run profile support to execute Skaffold itself and show logs in the console view
- support to properly stop Skaffold in dev mode (SIGINT instead of kill)
- UI editor.

ivanporty:
Closed via #56.

Issue #12, opened 2018-09-17
Create Skaffold run configuration and UI for single deploy.


ivanporty:
Run configuration allows a user to run Skaffold one time (deploy mode, `skaffold run`), and optionally, stream the logs into the console window.
Includes:
- declaration of run configuration
- factory to create configuration
- settings
- run profile support to execute Skaffold itself and show logs in the console view
- UI editor.

PR #45, opened 2018-10-03
Initializes Skaffold base UI for both configurations with drop down of all Skaffold files

ivanporty:
Progress towards #12 , #13.

Adds combo box and a base settings editor which will be used by both single run and continuous dev modes. Skaffold files from the whole project is collected and shown for the selection.

skaffold-combo-1

Follow up PRs will add necessary test infrastructure for mocks and mock services to create unit tests for the base settings editor.


ivanporty:
@tiembo @keyboardsurfer PTAL Kotlin style and idioms please. Thanks!

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Base settings editor for both Skaffold single run and continunous run configurations. Includes
+ * drop-down list of all Skaffold configuration files ([SkaffoldFilesComboBox]) found
+ * in the project and basic validation of the currently selected Skaffold file.
+ */
+open class BaseSkaffoldSettingsEditor : SettingsEditor<RunConfiguration>() {
+    private lateinit var skaffoldFilesComboBox: SkaffoldFilesComboBox
+
+    protected lateinit var basePanel: JPanel
```
etanshaul: why is this not private?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    override fun createEditor(): JComponent {
+        skaffoldFilesComboBox = SkaffoldFilesComboBox()
+        basePanel = panel {
```
etanshaul: I'm wondering if we can adopt some other pattern for declaring UI components so that they stand out more. Either in separate files or as class level variables (or something else)?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldDevSettingsEditor.kt
```diff
@@ -16,22 +16,9 @@
...
-class SkaffoldDevSettingsEditor : SettingsEditor<RunConfiguration>() {
-    override fun resetEditorFrom(s: RunConfiguration) {
-    }
-
-    override fun createEditor(): JComponent = JPanel()
-
-    override fun applyEditorTo(s: RunConfiguration) {
-    }
-}
+class SkaffoldDevSettingsEditor : BaseSkaffoldSettingsEditor()
```
etanshaul: is this still needed now that it has no implementation?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
```diff
@@ -0,0 +1,61 @@
...
+            row(message("skaffold.configuration.label")) { skaffoldFilesComboBox(grow) }
+        }
+
+        return basePanel
+    }
+
+    override fun applyEditorTo(runConfig: RunConfiguration) {
+        val selectedSkaffoldFile: VirtualFile =
+            skaffoldFilesComboBox.getItemAt(skaffoldFilesComboBox.selectedIndex)
+                ?: throw ConfigurationException(message("skaffold.no.file.selected.error"))
```
etanshaul: can you remind me - these ConfigurationExceptions, do they block the run configuration from being saved?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBox.kt
```diff
@@ -0,0 +1,76 @@
...
+ */
+class SkaffoldFilesComboBox : JComboBox<VirtualFile>() {
+    private lateinit var skaffoldFilesMutableModel: MutableComboBoxModel<VirtualFile>
+
+    /**
+     * Sets project for the combo box, and fills combo box with all Skaffold files available
+     * in this project. First file is selected by default.
+     */
+    fun setProject(project: Project) {
+        setRenderer(VirtualFileRenderer(project.baseDir))
```
etanshaul: can you just do `renderer =` here? Same as how you are doing for model (and Kotlin implicitly calls the Java setter).

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBox.kt
```diff
@@ -0,0 +1,76 @@
...
+            DefaultComboBoxModel<VirtualFile>(findSkaffoldFiles(project).toTypedArray())
+        model = skaffoldFilesMutableModel
+        if (model.size > 0) selectedIndex = 0
+    }
+
+    /**
+     * Sets given file as a selected Skaffold file in the drop-down. If model does not contain
+     * the file, adds it and selects it.
+     */
+    fun setSelectedSkaffoldFile(skaffoldFile: VirtualFile) {
```
etanshaul: I guess you'll consume this in future PRs?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
ivanporty: Hmm, I guess you're right, they should stand out a bit more. How about placing them as a separate function in the same file, named with a specific prefix/suffix, i.e. `create...UiPanel`, so for this case `fun createBaseSkaffoldSettingsEditorUiPanel()` then return resulting panel.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldDevSettingsEditor.kt
ivanporty: We will expand this for single run/continuous run options according to design docs, i.e. tail logs for single run etc.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBox.kt
ivanporty: Right, to select persisted file from settings.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBox.kt
ivanporty: Definitely, done. I wonder why IDEA didn't prompt me to, although it did on `model` access.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
ivanporty: I was planning to access this from child classes to expand UI, but I think it would be enough to just get base panel using `super` method. I don't think we need this field at all as of now. Removed.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBox.kt
etanshaul: thats interesting and slightly concerning that it didn't prompt you. It seems like it has the same behavior (by stepping into render it takes you to the setter). I wonder why it behaved differently.

ivanporty: Reverted - didn't work properly. `setRenderer` calls Swing property listeners, this might be the cause -  I guess Kotlin style properties are only when you assign the value. Interesting to hear Kotlin reviewers opinion.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
etanshaul: Thats good.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBox.kt
etanshaul: hmm but I thought that if you call a Java class from Kotlin using the `render = `, then it will invoke the Java setter for render. I'm not sure I understand how render is different from model (the setter does a bunch of side-effect stuff there too). 

would be good to understand this (I probably just need to read about it again).

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
```diff
@@ -0,0 +1,58 @@
...
+ * Base settings editor for both Skaffold single run and continunous run configurations. Includes
+ * drop-down list of all Skaffold configuration files ([SkaffoldFilesComboBox]) found
+ * in the project and basic validation of the currently selected Skaffold file.
+ */
+open class BaseSkaffoldSettingsEditor : SettingsEditor<RunConfiguration>() {
+    private lateinit var skaffoldFilesComboBox: SkaffoldFilesComboBox
+
+    override fun createEditor(): JComponent {
+        skaffoldFilesComboBox = SkaffoldFilesComboBox()
+        val basePanel = panel {
```
etanshaul: I'm ok with this for now. Once the UI gets bigger we can revisit.

etanshaul: you could even return this directly I think instead of saving a variable

keyboardsurfer:
Kotlin LGTM.

etanshaul:
thanks @keyboardsurfer

Issue #19, opened 2018-09-19
Create unit test infrastructure


ivanporty:
Create ContainerToolsRule similar to what CloudToolsRule has but probably faster and simpler, with better accomodation for Kotlin test framework .
The test rule should support:
- test services
- easy mocks
- test projects via heavyweight fixture

Conceivably we can follow what this blog suggests (https://blog.philipphauer.de/best-practices-unit-testing-kotlin/) and also consult Kotlin group on this.

ivanporty:
I think we can mark this as fixed now as we have most of the required support in place.

PR #46, opened 2018-10-05
Created initial support for Mockk, IJ test services, Skaffold files as a service to test SkaffoldComboBox.

ivanporty:
Part of #19.
Work towards #12, #13.

In order to thouroughly test `SkaffoldComboBox` class and mock Skaffold files search across the project, Skaffold file search is now made an IDE service. 

To support testing/mocking it with combo box and for later unit tests, test service support (with annotation, similar to Cloud Tools approach) is added, along with support for Mockk mocking library, which seems to be the most popular choice for Kotlin unit tests as of now.

### DIFF
--- a/common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
+++ b/common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
@@ -37,6 +37,8 @@ import kotlin.reflect.jvm.isAccessible
  *
  *  * Creates an [IdeaProjectTestFixture] and makes it available for tests via property. By default
  *    this text fixture is "light", i.e. instance of [LightIdeaTestFixture].
+ *  * Replaces all fields annotated with [TestService] in the
+ *    [PicoContainer][org.picocontainer.PicoContainer] with the field values.
  */
 class ContainerToolsRule(private val testInstance: Any) : TestRule {
     lateinit var ideaProjectTestFixture: IdeaProjectTestFixture
--- a/common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
+++ b/common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
@@ -16,13 +16,20 @@
 
 package com.google.container.tools.test
 
+import com.intellij.openapi.application.ApplicationManager
 import com.intellij.testFramework.EdtTestUtil
 import com.intellij.testFramework.fixtures.IdeaProjectTestFixture
 import com.intellij.testFramework.fixtures.IdeaTestFixtureFactory
 import com.intellij.util.ThrowableRunnable
+import io.mockk.MockKAnnotations
 import org.junit.rules.TestRule
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
+import org.picocontainer.MutablePicoContainer
+import kotlin.reflect.KClass
+import kotlin.reflect.KProperty1
+import kotlin.reflect.full.declaredMemberProperties
+import kotlin.reflect.jvm.isAccessible
 
 /**
  * A custom [TestRule] for Container Tools unit tests.
@@ -32,7 +39,7 @@ import org.junit.runners.model.Statement
  *  * Creates an [IdeaProjectTestFixture] and makes it available for tests via property. By default
  *    this text fixture is "light", i.e. instance of [LightIdeaTestFixture].
  */
-class ContainerToolsRule : TestRule {
+class ContainerToolsRule(val testInstance: Any) : TestRule {
     lateinit var ideaProjectTestFixture: IdeaProjectTestFixture
 
     override fun apply(baseStatement: Statement, description: Description): Statement =
@@ -51,9 +58,40 @@ class ContainerToolsRule : TestRule {
         ideaProjectTestFixture =
             IdeaTestFixtureFactory.getFixtureFactory().createLightFixtureBuilder().fixture
         EdtTestUtil.runInEdtAndWait(ThrowableRunnable { ideaProjectTestFixture.setUp() })
+
+        MockKAnnotations.init(testInstance, relaxUnitFun = true)
+        replaceServices()
     }
 
     private fun tearDownRule() {
         EdtTestUtil.runInEdtAndWait(ThrowableRunnable { ideaProjectTestFixture.tearDown() })
+    }
+
+    /**
+     * Replaces all services annotated with [TestService].
+     */
+    private fun replaceServices() {
+        for (member in testInstance::class.declaredMemberProperties) {
+            if (member.annotations.filter { it is TestService }.isNotEmpty()) {
+                member as KProperty1<Any?, Any?>
+                member.isAccessible = true
+                val service: Any = member.get(testInstance)!!
+                setService(member::class, service)
+            }
+        }
+    }
+
+    /**
+     * Replaces the service binding in the [MutablePicoContainer] with the given instance and
+     * returns the original service instance.
+     *
+     * @param clazz the class of the registered service
+     * @param newInstance the new instance to register
+     */
+    private fun setService(clazz: KClass<*>, newInstance: Any) {
+        val applicationContainer =
+            ApplicationManager.getApplication().picoContainer as MutablePicoContainer
+        applicationContainer.unregisterComponent(clazz.simpleName)
+        applicationContainer.registerComponentInstance(clazz.simpleName, newInstance)
     }
 }
--- a/common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
+++ b/common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
@@ -26,7 +26,6 @@ import org.junit.rules.TestRule
 import org.junit.runner.Description
 import org.junit.runners.model.Statement
 import org.picocontainer.MutablePicoContainer
-import kotlin.reflect.KClass
 import kotlin.reflect.KProperty1
 import kotlin.reflect.full.declaredMemberProperties
 import kotlin.reflect.jvm.isAccessible
@@ -39,7 +38,7 @@ import kotlin.reflect.jvm.isAccessible
  *  * Creates an [IdeaProjectTestFixture] and makes it available for tests via property. By default
  *    this text fixture is "light", i.e. instance of [LightIdeaTestFixture].
  */
-class ContainerToolsRule(val testInstance: Any) : TestRule {
+class ContainerToolsRule(private val testInstance: Any) : TestRule {
     lateinit var ideaProjectTestFixture: IdeaProjectTestFixture
 
     override fun apply(baseStatement: Statement, description: Description): Statement =
@@ -76,7 +75,7 @@ class ContainerToolsRule(val testInstance: Any) : TestRule {
                 member as KProperty1<Any?, Any?>
                 member.isAccessible = true
                 val service: Any = member.get(testInstance)!!
-                setService(member::class, service)
+                setService(service)
             }
         }
     }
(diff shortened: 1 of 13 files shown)


ivanporty:
@tiembo @keyboardsurfer PTAL Kotlin style and idioms please. What do you think on unit test style and approach? Thanks!

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
```diff
@@ -32,7 +38,7 @@ import org.junit.runners.model.Statement
  *  * Creates an [IdeaProjectTestFixture] and makes it available for tests via property. By default
  *    this text fixture is "light", i.e. instance of [LightIdeaTestFixture].
  */
-class ContainerToolsRule : TestRule {
+class ContainerToolsRule(private val testInstance: Any) : TestRule {
```
keyboardsurfer: Can `testInstance` scoped to something else than `Any`?

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
```diff
@@ -51,9 +57,42 @@ class ContainerToolsRule : TestRule {
...
+
+    /**
+     * Replaces the service binding in the [MutablePicoContainer] with the given instance and
+     * returns the original service instance.
+     *
+     * @param newInstance the new instance to register
+     */
+    private fun setService(newInstance: Any) {
+        val applicationContainer =
+            ApplicationManager.getApplication().picoContainer as MutablePicoContainer
```
keyboardsurfer: `with (Application...) { ... }` is more concise

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -50,13 +51,20 @@ fun isSkaffoldFile(file: VirtualFile): Boolean {
...
- * Finds all Skaffold configuration YAML files in the given project.
- *
- * @param project IDE project to search Skaffold file in
- * @return List of Skaffold configuration files in the project.
- */
-fun findSkaffoldFiles(project: Project): List<VirtualFile> {
-    return FileTypeIndex.getFiles(YAMLFileType.YML, GlobalSearchScope.allScope(project))
-        .filter { isSkaffoldFile(it) }
+/** Obtains current active implementation of [SkaffoldFileService] */
+fun getSkaffoldFileService() = ServiceManager.getService(SkaffoldFileService::class.java)!!
```
keyboardsurfer: How likely is this service going to be `null`?

etanshaul: instead of a `getSkaffoldFileService` here, why not a `SkaffoldFileServce#getInstance`?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
```diff
@@ -50,13 +51,20 @@ fun isSkaffoldFile(file: VirtualFile): Boolean {
...
+
+/** IDE service for findind Skaffold files in the given IDE project. */
+class SkaffoldFileService {
```
etanshaul: how did you decide which functions moved to here vs staying in `SkaffoldFiles.kt`? seems like there isn't a clear boundary now.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBox.kt
```diff
@@ -51,9 +53,23 @@ class SkaffoldFilesComboBox : JComboBox<VirtualFile>() {
      * the file, adds it and selects it.
      */
     fun setSelectedSkaffoldFile(skaffoldFile: VirtualFile) {
-        skaffoldFilesMutableModel.addElement(skaffoldFile)
+        var existingElement = false
+        for (i in 0..skaffoldFilesMutableModel.size) {
+            if (skaffoldFilesMutableModel.getElementAt(i) == skaffoldFile) {
```
etanshaul: does VirtualFile implement `equals`?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBoxTest.kt
```diff
@@ -0,0 +1,117 @@
...
+import org.junit.Test
+
+/** Unit tests for [SkaffoldFilesComboBox] */
+class SkaffoldFilesComboBoxTest {
+    @get:Rule
+    val containerToolsRule = ContainerToolsRule(this)
+
+    @MockK
+    @TestService
+    lateinit var mockSkaffoldFileService: SkaffoldFileService
```
etanshaul: private?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBoxTest.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    private lateinit var skaffoldFilesComboBox: SkaffoldFilesComboBox
+
+    @Before
+    fun setUp() {
+        skaffoldFilesComboBox = SkaffoldFilesComboBox()
+    }
+
+    @Test
+    fun `empty combobox for empty project has no files`() {
```
etanshaul: maybe clarify the wording a bit? I'm not sure what the "for empty project" part means?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBoxTest.kt
```diff
@@ -0,0 +1,117 @@
...
+        skaffoldFilesComboBox.setProject(project)
+        skaffoldFilesComboBox.setSelectedSkaffoldFile(mockSkaffoldFile2)
+
+        // make sure the file is not double-added.
+        assertThat(skaffoldFilesComboBox.model.size).isEqualTo(2)
+        assertThat(skaffoldFilesComboBox.getSelectedSkaffoldFile()).isEqualTo(mockSkaffoldFile2)
+    }
+
+    @Test
+    fun `setSelectedSkaffoldFile with non-existing file changes selection and adds element`() {
```
etanshaul: Why do we want this behavior? Seems to me that if you try to set a non-existent skaffold file then it should be an error.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
ivanporty: Not likely in production, but I added that to make sure tests are working fine and I'm replacing the test service right. If it's not likely to be null, do you recommend avoid `!!`?

ivanporty: If I understand, there are no static functions in Kotlin, so this call needs to be wrapped into companion object somehow, and I used the plain function instead. I didn't use `getInstance()` to avoid confusion. Interesting question though, @keyboardsurfer, what is your opinion? Is it better to use companion object and get the single instance of this service or use of the separate function is allowed as well?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBoxTest.kt
ivanporty: sounds good, let me reword this a bit.

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBoxTest.kt
ivanporty: For existing run configurations where Skaffold file was set. If user opens them and that file was removed, I believe we still want this selection to show (and remind what was the configuration before), but the error will show as well saying we don't have this file in the project anymore.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
etanshaul: seems like a companion object is perfect for this use case, but I'll defer to the Kotlin experts

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBoxTest.kt
etanshaul: ah. yea makes sense.

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
ivanporty: Couldn't see anything more specific, it could be any type of object, we are only interested in its properties and annotations here. What would be a Kotlin way here?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldFilesComboBox.kt
ivanporty: A bit hard to find particular implementation, there are many of them, but the javadoc claims the following:
```
* A particular file is represented by equal {@code VirtualFile} instances for the entire lifetime of the IntelliJ IDEA process, unless the file
 * is deleted, in which case {@link #isValid()} will return {@code false}.
 * <p/>
 * VirtualFile instances are created on request, so there can be several instances corresponding to the same file.
 * All of them are equal, have the same hashCode and use shared storage for all related data, including user data (see {@link UserDataHolder}).
```
Last sentence says it should work.

etanshaul: cool

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFiles.kt
ivanporty: You're right, both of them fit nicely in the new service class. Let me move the other one and rename the file as well.

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
ivanporty: Thanks for the tip, done.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFileService.kt
```diff
@@ -0,0 +1,70 @@
...
+import java.util.regex.Pattern
+
+// see https://github.com/GoogleContainerTools/skaffold/blob/master/examples/annotated-skaffold.yaml
+private const val SKAFFOLD_API_HEADER_REGEX = """\s*apiVersion:\s*skaffold/v"""
+private val SKAFFOLD_API_HEADER_PATTERN: Pattern by lazy {
+    Pattern.compile(SKAFFOLD_API_HEADER_REGEX)
+}
+
+/** Obtains current active implementation of [SkaffoldFileService] */
+fun getSkaffoldFileService() = ServiceManager.getService(SkaffoldFileService::class.java)!!
```
etanshaul: Was looking at how Kotlin does companion objects. I think something like this _may_ work:

```
class SkaffoldFileService {

  companion object {
    val instance: SkaffoldFileService
      get() = ServiceManager.getService(SkaffoldFileService::class.java)!!
  }  

  fun someOtherFun() ...

}
```
then use it 
```
SkaffoldFileService.instance.someOtherFun()
```

Makes the calling code a bit cleaner. 

wdyt?

Note: I'm not really sure what the difference is between:
```
SkaffoldFileService.instance.someOtherFun()
```
and 
```
SkaffoldFileService.Companion.instance.someOtherFun()
```
as they both seem to compile and do the same thing

etanshaul: to answer my question at the end: I think the `.Companion` syntax is for calling Kotlin code from Java. In Kotlin, both work.

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
keyboardsurfer: If it could be anything then `Any` is appropriate.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFileService.kt
ivanporty: Looks good, I agree this look more familiar for us and it seems to be a completely normal practice to define single instance of the object. Changed to use companion object and instance field. Thanks!

PR #48, opened 2018-10-09
 Creates base config for both Skaffold run types to persist common Skaffold settings.

ivanporty:
Work towards #12, #13.
Part of #19.

Adds base run configuration (`BaseSkaffoldRunConfiguration`) which includes persisted attribute to store saved Skaffold configuration file between restarts and changes.
Includes test for base settings editor that uses this configuration settings.

For unit tests, introduces `@UiTest` annotation to allow individual tests to run on UI thread, if they manipulate or access UI components or related state to avoid possible unit test flakiness.

### DIFF
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
@@ -31,7 +31,7 @@ import javax.swing.JComponent
  * drop-down list of all Skaffold configuration files ([SkaffoldFilesComboBox]) found
  * in the project and basic validation of the currently selected Skaffold file.
  */
-open class BaseSkaffoldSettingsEditor : SettingsEditor<BaseSkaffoldRunConfiguration>() {
+open class BaseSkaffoldSettingsEditor : SettingsEditor<AbstractSkaffoldRunConfiguration>() {
     @VisibleForTesting
     lateinit var skaffoldFilesComboBox: SkaffoldFilesComboBox
 
@@ -44,7 +44,7 @@ open class BaseSkaffoldSettingsEditor : SettingsEditor<BaseSkaffoldRunConfigurat
         return basePanel
     }
 
-    override fun applyEditorTo(runConfig: BaseSkaffoldRunConfiguration) {
+    override fun applyEditorTo(runConfig: AbstractSkaffoldRunConfiguration) {
         val selectedSkaffoldFile: VirtualFile =
             skaffoldFilesComboBox.getItemAt(skaffoldFilesComboBox.selectedIndex)
                 ?: throw ConfigurationException(message("skaffold.no.file.selected.error"))
@@ -57,7 +57,7 @@ open class BaseSkaffoldSettingsEditor : SettingsEditor<BaseSkaffoldRunConfigurat
         runConfig.skaffoldConfigurationFilePath = selectedSkaffoldFile.path
     }
 
-    override fun resetEditorFrom(runConfig: BaseSkaffoldRunConfiguration) {
+    override fun resetEditorFrom(runConfig: AbstractSkaffoldRunConfiguration) {
         skaffoldFilesComboBox.setProject(runConfig.project)
         runConfig.skaffoldConfigurationFilePath?.let {
             LocalFileSystem.getInstance().findFileByPath(it)
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
@@ -18,9 +18,9 @@ package com.google.container.tools.skaffold.run
 
 import com.google.container.tools.skaffold.SkaffoldFileService
 import com.google.container.tools.skaffold.message
-import com.intellij.execution.configurations.RunConfiguration
 import com.intellij.openapi.options.ConfigurationException
 import com.intellij.openapi.options.SettingsEditor
+import com.intellij.openapi.vfs.LocalFileSystem
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.ui.layout.panel
 import javax.swing.JComponent
@@ -30,7 +30,7 @@ import javax.swing.JComponent
  * drop-down list of all Skaffold configuration files ([SkaffoldFilesComboBox]) found
  * in the project and basic validation of the currently selected Skaffold file.
  */
-open class BaseSkaffoldSettingsEditor : SettingsEditor<RunConfiguration>() {
+open class BaseSkaffoldSettingsEditor : SettingsEditor<BaseSkaffoldRunConfiguration>() {
     private lateinit var skaffoldFilesComboBox: SkaffoldFilesComboBox
 
     override fun createEditor(): JComponent {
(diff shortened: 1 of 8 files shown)


ivanporty:
@keyboardsurfer or @tiembo PTAL Kotlin style and idioms please. Thanks!

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditorTest.kt
```diff
@@ -0,0 +1,102 @@
...
+        })
+
+        // default is project with no Skaffold files
+        every { mockSkaffoldFileService.findSkaffoldFiles(any()) } returns listOf()
+        // return test fixture project for settings
+        every { mockSkaffoldSettings.project } answers
+            { containerToolsRule.ideaProjectTestFixture.project }
+    }
+
+    @Test(expected = ConfigurationException::class)
```
etanshaul: caught my attention on initial scan. See [http://go/java-testing/exceptions#discouraged](http://go/java-testing/exceptions#discouraged).

ivanporty: You're right, I tried to find something in Mockk to handle them better but couldn't. changed to assert

ivanporty: Changed to use test helper method `expectThrows` similar to what we use in Cloud Tools, since IDE has Junit4 bundled which doesn't have built-in `assertThrows` method.

etanshaul: Looks good. Reviewing the rest

etanshaul:
windows CI failed. but it looks like a transient failuure

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -52,11 +53,26 @@ class SkaffoldDevConfiguration(
...
 
     override fun getConfigurationEditor(): SettingsEditor<out RunConfiguration> =
         SkaffoldDevSettingsEditor()
 }
+
+/**
+ * Base class for Skaffold run configurations, includes base properties such as Skaffold
+ * configuration file path.
+ */
+abstract class BaseSkaffoldRunConfiguration(
```
etanshaul: Instead of `BaseSkaffoldRunConfiguration`, how about just `SkaffoldRunConfiguration`?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -52,11 +53,26 @@ class SkaffoldDevConfiguration(
...
+    project: Project,
+    factory: ConfigurationFactory,
+    name: String
+) : RunConfigurationBase(project, factory, name) {
+
+    /** Persisted Skaffold config file name for Skaffold run configurations. */
```
etanshaul: is this a relative path / absolute path? maybe make it clear in the doc

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
```diff
@@ -42,17 +44,23 @@ open class BaseSkaffoldSettingsEditor : SettingsEditor<RunConfiguration>() {
...
         }
+
+        // save properties
+        runConfig.skaffoldConfigurationFilePath = selectedSkaffoldFile.path
     }
 
-    override fun resetEditorFrom(runConfig: RunConfiguration) {
+    override fun resetEditorFrom(runConfig: BaseSkaffoldRunConfiguration) {
         skaffoldFilesComboBox.setProject(runConfig.project)
+        runConfig.skaffoldConfigurationFilePath?.let {
```
etanshaul: just to make sure I understand this snippet:

if there is a skaffold path set, you locate the corresponding `File` on the filesystem, then set the file in the combobox model?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditorTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        // default is project with no Skaffold files
+        every { mockSkaffoldFileService.findSkaffoldFiles(any()) } returns listOf()
+        // return test fixture project for settings
+        every { mockSkaffoldSettings.project } answers
+            { containerToolsRule.ideaProjectTestFixture.project }
+    }
+
+    @Test
+    @UiTest
+    fun `settings are invalid for project with no existing Skaffold files`() {
```
etanshaul: regarding these freeform function names: I like this is a lot, but maybe lets still give it some structure so that we know exactly what to expect from the tests when reading it:

i.e. in Java we did something like 

`resetFrom_whenNoSkaffoldFilesFound_throwsConfigurationException`

This makes it very clear the behavior and expectation.

For Kotlin, if we structured it similarly we could have something like (feel free to deviate from this):

`when project has no Skaffold files a ConfigurationException is thrown`

wdyt?

etanshaul:
PR looks great. just a few small comments / suggestions

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
ivanporty: Correct, regardless of whether this file exists or not in the filesystem anymore, its path is taken, converted into VirtualFile and shown in the combo box.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
ivanporty: absolute, made an update and linked to a source of the path.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
ivanporty: I agree `Base` maybe a bit misleading, but just saying `SkaffoldRunConfiguration` may imply this configuration is used as is which is not quite true, since we intend to keep common settings here. How about `CommonSkaffoldRunConfiguration`? or maybe `Generic`

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditorTest.kt
ivanporty: Sounds like a good idea, with this freeform names I sometimes deviate a bit, this simple rule should help having clearer names. Excellent suggestion!

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
etanshaul: or we can just be blunt about it and say `AbstractSkaffoldRunConfiguration`?

ivanporty: Went with just `SkaffoldRunConfiguration`, after giving it another thought, it should be good enough.

ivanporty: Although it is indeed abstract, let's do `AbstractSkaffoldRunConfiguration` instead if you don't object.

PR #49, opened 2018-10-15
 Adds Skaffold executor service to build Skaffold commands and execute process.

ivanporty:
Progress towards #12, #13.

Creates a service that constructs Skaffold command line and launches system process based on the resulting command line. Service is defined with an interface. Default implementation assumes Skaffold is already installed and present in PATH on the system (current behavior for command line users.)

Next PR will use this service to start processes based on run configurations.
The service interface is also going to be used when Skaffold managed installation implementation is complete and will be used to execute Skaffold using managed binary.

### DIFF
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
@@ -16,6 +16,23 @@
 
 package com.google.container.tools.skaffold
 
-class SkaffoldExecutorService {
-    
+class DefaultSkaffoldExecutorService : SkaffoldExecutorService {
+    override var skaffoldConfigurationFilePath: String?
+        get() = TODO("not implemented") //To change initializer of created properties use File | Settings | File Templates.
+        set(value) {}
+
+    override fun executeSingleRun() {
+        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
+    }
+
+    override fun executeDevMode() {
+        TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
+    }
+}
+
+interface SkaffoldExecutorService {
+    var skaffoldConfigurationFilePath: String?
+
+    fun executeSingleRun()
+    fun executeDevMode()
 }
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
@@ -17,21 +17,27 @@
 package com.google.container.tools.skaffold
 
 import com.google.common.annotations.VisibleForTesting
+import com.google.container.tools.skaffold.SkaffoldExecutorSettings.ExecutionMode
 import com.intellij.openapi.components.ServiceManager
 import java.io.File
 import java.nio.file.Path
 import java.nio.file.Paths
 
 /**
- * Interface for Skaffold execution service. This service builds and launches Skaffold process from
- * the given set of standard flags and settings. [DefaultSkaffoldExecutorService] assumes Skaffold
- * to be already installed and available.
+ * Abstract implementation for Skaffold execution service. This service builds and launches Skaffold
+ * process from the given set of standard flags and settings.
+ * Constructs Skaffold command line and build resulting process.
+ * Path to Skaffold executable must be set by subclasses.
+ * [DefaultSkaffoldExecutorService] assumes Skaffold to be already installed and available.
  */
-interface SkaffoldExecutorService {
+abstract class SkaffoldExecutorService {
     companion object {
         val instance
             get() = ServiceManager.getService(SkaffoldExecutorService::class.java)!!
     }
+
+    /** Path for Skaffold executable, any form supported by [ProcessBuilder] */
+    protected abstract var skaffoldExecutablePath: Path
 
     /**
      * Creates Skaffold command line from the given settings and returns resulting launched
@@ -40,50 +46,8 @@ interface SkaffoldExecutorService {
      * @param settings Settings with flags supported by Skaffold.
      * @return [SkaffoldProcess] with resulting process and the command line used to launch it.
      */
-    fun executeSkaffold(settings: SkaffoldExecutorSettings): SkaffoldProcess
-
-    /**
-     * Execution mode for Skaffold, single run, continuous development, etc.
-     */
-    enum class SkaffoldExecutionMode(val modeFlag: String) {
-        SINGLE_RUN("run"),
-        DEV("dev")
-    }
-
-    /**
-     * Set of settings to control Skaffold execution, including flags and execution mode.
-     *
-     * @property executionMode Mandatory execution mode for Skaffold, see [SkaffoldExecutionMode].
-     * @property skaffoldConfigurationFilePath Optional, location of the Skaffold YAML
-     *           configuration file. If not provided, default `skaffold.yaml` used.
-     * @property workingDirectory Optional, working directory where Skaffold needs to be launched.
-     *           This is usually set to project working directory.
-     */
-    data class SkaffoldExecutorSettings(
-        val executionMode: SkaffoldExecutionMode,
-        val skaffoldConfigurationFilePath: String? = null,
-        var workingDirectory: File? = null
-    )
-
-    /**
-     * Data object with launched Skaffold process and its command line.
-     *
-     * @property process System process for Skaffold.
-     * @property commandLine Command line used to launch the process.
-     */
-    data class SkaffoldProcess(val process: Process, val commandLine: String)
-}
-
-/**
- * Abstract implementation of [SkaffoldExecutorService], constructs Skaffold command line and
- * build resulting process. Path to Skaffold executable must be set by subclasses.
- */
-abstract class AbstractSkaffoldExecutorService : SkaffoldExecutorService {
-    /** Path for Skaffold executable, any form supported by [ProcessBuilder] */
-    protected abstract var skaffoldExecutablePath: Path
-
-    override fun executeSkaffold(settings: SkaffoldExecutorService.SkaffoldExecutorSettings):
-        SkaffoldExecutorService.SkaffoldProcess {
+    fun executeSkaffold(settings: SkaffoldExecutorSettings):
+        SkaffoldProcess {
         val commandList = mutableListOf<String>()
         with(commandList) {
             add(skaffoldExecutablePath.toString())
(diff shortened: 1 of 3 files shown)


ivanporty:
@keyboardsurfer or @tiembo PTAL Kotlin style and idioms please. Thanks!

etanshaul:
@ivanporty if my review on this is blocking you feel free branch off of this and continue put out other PRs

ivanporty:
@etanshaul Sounds like a plan, other work is in progress in parallel.

etanshaul:
great. thanks!

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -0,0 +1,125 @@
...
+     *
+     * @param settings Settings with flags supported by Skaffold.
+     * @return [SkaffoldProcess] with resulting process and the command line used to launch it.
+     */
+    fun executeSkaffold(settings: SkaffoldExecutorSettings): SkaffoldProcess
+
+    /**
+     * Execution mode for Skaffold, single run, continuous development, etc.
+     */
+    enum class SkaffoldExecutionMode(val modeFlag: String) {
```
etanshaul: I'm not sure the answer but would it be better if this lived in the data class below?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Interface for Skaffold execution service. This service builds and launches Skaffold process from
+ * the given set of standard flags and settings. [DefaultSkaffoldExecutorService] assumes Skaffold
+ * to be already installed and available.
+ */
+interface SkaffoldExecutorService {
+    companion object {
+        val instance
+            get() = ServiceManager.getService(SkaffoldExecutorService::class.java)!!
```
keyboardsurfer: Can we get rid of the `!!`?

keyboardsurfer: Also, do we need to call getService every time the property is accessed?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -0,0 +1,125 @@
...
+     */
+    enum class SkaffoldExecutionMode(val modeFlag: String) {
+        SINGLE_RUN("run"),
+        DEV("dev")
+    }
+
+    /**
+     * Set of settings to control Skaffold execution, including flags and execution mode.
+     */
+    data class SkaffoldExecutorSettings(
```
keyboardsurfer: This documentation style makes it hard to see the actual code. Can you pull it up into @param values or have an explicit `constructor(...)?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
ivanporty: This is a call to underlying IDE API, which could potentially return `null`, but that would indicate a configuration error that should never happen, hence `!!`. `getService()` is cached itself and the service can be changed dynamically, so we always call it as a getter. Instead of `!!`, do you recommend just declaring the `instance` as non-null? What would happen if `getService()` returns null?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
ivanporty: Sounds good, thanks, changed.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
ivanporty: I don't have a very strong feeling, both approaches seems fine to me, the only thing it's going to be less visible and the name is going to be very long and deeply nested, making it harder to read. wdyt?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
etanshaul: > Also, do we need to call getService every time the property is accessed?

So does that mean that even though its a `val` it will get called each time? I thought it would only get called once? @ivanporty something to look into

etanshaul: or perhaps I misread the comment and @keyboardsurfer was just checking to make sure we didn't actually need to get it fresh each time.

ivanporty: we will change that later a bit as well - when we have another implementation of executor available (the one using managed Skaffold installation), so we will choose implementation dynamically. So eventually I think it's a good idea to always get a current implementation using `ServiceManager`. Do we have any concerns about performance here?

keyboardsurfer: @ivanporty: Your service and potentially whole application would crash with an NPE if that'd return null.
[See doc](https://kotlinlang.org/docs/reference/null-safety.html#the--operator)

Calls to a custom getter are not cached. This means that your results can change over time, depending on the ServiceManager.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -0,0 +1,122 @@
...
+     * @property commandLine Command line used to launch the process.
+     */
+    data class SkaffoldProcess(val process: Process, val commandLine: String)
+}
+
+/**
+ * Abstract implementation of [SkaffoldExecutorService], constructs Skaffold command line and
+ * build resulting process. Path to Skaffold executable must be set by subclasses.
+ */
+abstract class AbstractSkaffoldExecutorService : SkaffoldExecutorService {
```
etanshaul: Kotlin interfaces can contain method implementations. Given that, is the abstract class still needed or can you push things up to the interface?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
etanshaul: I think it makes more sense to be part of the settings

regarding the nested names, maybe promote SkaffoldExecutionSettings to its own class?

etanshaul: if you prefer we can refactor this separately from this PR

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
etanshaul: Looking at this again, I think it makes sense to have this as a custom getter and defer the fetching of the service to the `ServiceManager` (which internally itself caches the service).

Regarding the `!!` and potential NPE, just an idea:

What if we created an extension function on `ServiceManager` that would check get the service and check if its null. If its null, then throw a new exception (something like `ApplicationServiceNotDefinedException` with a message). Its ok for this to bubble up and be an IDE exception in this case since we are the once defining this service, so it should really never be null (although we do still need to be very careful when loading across plugin.xml boundaries).

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+/**
+ * Default implementation of Skaffold execution service, where Skaffold executable is assumed to
+ * be already installed on the system and be available in PATH.
+ */
+class DefaultSkaffoldExecutorService : AbstractSkaffoldExecutorService() {
+    // use executable available in PATH
+    override var skaffoldExecutablePath: Path = Paths.get("skaffold")
```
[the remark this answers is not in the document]
ivanporty: Good option. Hard to foresee anything except default / managed Skaffold options as of now, but if we add something else (like pure `kubectl` or some other deployment helper tools), this may get messy and subclassing might be handy. I agree, let's leave this  as it is and keep this option in mind.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
ivanporty: It feels like interface declares just an intention, and all needed data structures like `SkaffoldExecutorSettings`, so it's pretty full of information as it is, and abstract class describes the actual execution details. We can move that to interface, but it may make things a bit cluttered. wdyt?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
ivanporty: Good idea, let's see how we use this in the next PR, and if this feels better to have it as a separate class and move enum inside of it, we'll refactor it.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
etanshaul: oh right.... that sucks that you can't do this for static methods.

The separate function feels a lot less clean so I'm inclined to just leave it as is and let it blow up. After all, is this any different than the behavior in our other (java) plugin?

ivanporty: Right, we would otherwise need to check for `null` all over the place. Missing our own service would indicate a serious error, so it would be same as in the Cloud Tools plugin.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
ivanporty: I agree, simplified a bit. PTAL

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
etanshaul: yeah. lets just leave it then. wdyt?

PR #53, opened 2018-10-25
Creates new skaffold-editing module and live template

etanshaul:
-  Creates a new `skaffold-editing` module to house the editing generation support
- Sets up a new live template for skaffold configs. The config is currently hardcoded in the `Skaffold.xml` live template config file. As is, we would need to update it to keep up with Skaffold version / api changes. We can look into enhancing this to be more dynamic in subsequent PRs.

How it works in the IDE:

user types in the live template keyword as usual (in this case its `skaffold`):
image

selecting the template generates a starter config:
image

### DIFF
--- a/skaffold-editing/src/main/kotlin/com/google/container/tools/skaffold/editing/SkaffoldContextType.kt
+++ b/skaffold-editing/src/main/kotlin/com/google/container/tools/skaffold/editing/SkaffoldContextType.kt
@@ -1,0 +1,25 @@
+package com.google.container.tools.skaffold.editing
+
+import com.intellij.codeInsight.template.TemplateContextType
+import com.intellij.psi.PsiFile
+
+private const val SKAFFOLD_TEMPLATE_ID = "SKAFFOLD"
+private const val SKAFFOLD_TEMPLATE_NAME = "Skaffold"
+
+/**
+ * Defines the [TemplateContextType] for Skaffold files.
+ */
+class SkaffoldContextType(
+    id: String = SKAFFOLD_TEMPLATE_ID,
+    presentableName: String = SKAFFOLD_TEMPLATE_NAME
+) :
+    TemplateContextType(id, presentableName) {
+
+    /**
+     * A file is in context for Skaffold live templates if it is a yaml file.
+     */
+    override fun isInContext(file: PsiFile, offset: Int): Boolean {
+        return file.name.endsWith(suffix = "yaml", ignoreCase = true)
+            || file.name.endsWith(suffix = "yml", ignoreCase = true)
+    }
+}
--- a/skaffold-editing/src/main/kotlin/com/google/container/tools/skaffold/editing/SkaffoldLiveTemplatesProvider.kt
+++ b/skaffold-editing/src/main/kotlin/com/google/container/tools/skaffold/editing/SkaffoldLiveTemplatesProvider.kt
@@ -1,0 +1,20 @@
+package com.google.container.tools.skaffold.editing
+
+import com.intellij.codeInsight.template.impl.DefaultLiveTemplatesProvider
+
+private const val SKAFFOLD_LIVE_TEMPLATE_PATH = "liveTemplates/Skaffold"
+
+/**
+ * A Skaffold [DefaultLiveTemplatesProvider] that provides the live templates used for generating /
+ * editing Skaffold configuration files.
+ */
+class SkaffoldLiveTemplatesProvider : DefaultLiveTemplatesProvider {
+
+    override fun getDefaultLiveTemplateFiles(): Array<String> {
+        return arrayOf(SKAFFOLD_LIVE_TEMPLATE_PATH)
+    }
+
+    override fun getHiddenLiveTemplateFiles(): Array<String>? {
+        return emptyArray()
+    }
+}
--- a/skaffold-editing/src/main/resources/liveTemplates/Skaffold.xml
+++ b/skaffold-editing/src/main/resources/liveTemplates/Skaffold.xml
@@ -1,0 +1,8 @@
+<templateSet group="Skaffold">
+  <template name="skaffold" value="apiVersion: apps/v1alpha4&#10;kind: Config&#10;build:&#10;  artifacts:&#10;    - image: $IMAGE$ &#10;deploy:&#10;  kubectl:&#10;      manifests:&#10;     - $MANIFEST$&#10;    &#10;" description="Skaffold Config" toReformat="true" toShortenFQNames="true">
+    <variable name="IMAGE" expression="" defaultValue="" alwaysStopAt="true" />
+    <context>
+      <option name="SKAFFOLD" value="true" />
+    </context>
+  </template>
+</templateSet>
--- a/.idea/compiler.xml
+++ b/.idea/compiler.xml
@@ -2,8 +2,12 @@
 <project version="4">
   <component name="CompilerConfiguration">
     <bytecodeTargetLevel>
+      <module name="common-test-lib_main" target="1.8" />
+      <module name="common-test-lib_test" target="1.8" />
       <module name="google-container-tools-intellij_main" target="1.8" />
       <module name="google-container-tools-intellij_test" target="1.8" />
+      <module name="skaffold-editing_main" target="1.8" />
+      <module name="skaffold-editing_test" target="1.8" />
       <module name="skaffold_main" target="1.8" />
       <module name="skaffold_test" target="1.8" />
     </bytecodeTargetLevel>
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -67,4 +67,5 @@ allprojects {
 
 dependencies {
     compile(project(":skaffold"))
+    compile(project(":skaffold-editing"))
 }
--- a/skaffold-editing/build.gradle.kts
+++ b/skaffold-editing/build.gradle.kts
@@ -14,6 +14,6 @@
  * limitations under the License.
  */
 
-include 'skaffold'
-include 'skaffold-editing'
-include 'common-test-lib'
+dependencies {
+    compile(project(":common-test-lib"))
+}
--- a/skaffold-editing/src/main/resources/META-INF/skaffold-editing.xml
+++ b/skaffold-editing/src/main/resources/META-INF/skaffold-editing.xml
@@ -1,0 +1,23 @@
+<!--
+  ~ Copyright 2018 Google Inc. All Rights Reserved.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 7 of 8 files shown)


etanshaul:
working on tests

etanshaul:
@ivanporty ready for review

etanshaul:
@balopat 

This PR creates a bare bones `skaffold.yaml` file for the user to get started via IntelliJ's [Live Template](https://www.jetbrains.com/help/idea/using-live-templates.html) mechanism (see the PR description for more details). 

Currently I have it creating the following:


I was wondering if you could comment on what a good getting started `skaffold.yaml` should look like based on best practices. For example, should it default to using GCB?

Note this is a simple context-free template. In the future once we have more context from the user's project we should be able to generate more tailored configs.

skaffold-editing/src/main/kotlin/com/google/container/tools/skaffold/editing/SkaffoldContextType.kt
```diff
@@ -0,0 +1,41 @@
...
+    id: String = SKAFFOLD_TEMPLATE_ID,
+    presentableName: String = SKAFFOLD_TEMPLATE_NAME
+) :
+    TemplateContextType(id, presentableName) {
+
+    /**
+     * A file is in context for Skaffold live templates if it is a yaml file.
+     */
+    override fun isInContext(file: PsiFile, offset: Int): Boolean {
+        return file.name.endsWith(suffix = "yaml", ignoreCase = true) ||
```
ivanporty: This seems to be a bit low-level for IDE APIs, how about using `getFileType()` instead, in the manner we used it for `findSkaffoldFiles()` function?

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
```diff
@@ -64,33 +72,80 @@ class ContainerToolsRule(private val testInstance: Any) : TestRule {
...
      * Replaces all services annotated with [TestService].
      */
     private fun replaceServices() {
-        for (member in testInstance::class.declaredMemberProperties) {
-            if (member.annotations.filter { it is TestService }.isNotEmpty()) {
-                member as KProperty1<Any?, Any?>
-                member.isAccessible = true
-                val service: Any = member.get(testInstance)!!
-                setService(service)
+        for (member in getMembersWithAnnotation(TestService::class.java)) {
```
ivanporty: Is there a reason why we need to use Java class here instead of native Kotlin class?

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
```diff
@@ -64,33 +72,80 @@ class ContainerToolsRule(private val testInstance: Any) : TestRule {
...
+        }
+    }
+
+    /**
+     * Creates all [File] annotated with [TestFile] in the given directory name.
+     *
+     * @param directoryName the name of the directory to create the test files in
+     */
+    private fun createTestFiles(directoryName: String) {
+        for (member in getMembersWithAnnotation(TestFile::class.java)) {
```
ivanporty: Same question about Java class.

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
```diff
@@ -64,33 +72,80 @@ class ContainerToolsRule(private val testInstance: Any) : TestRule {
...
+            member.setter.call(testInstance, file)
         }
     }
 
+    /**
+     * Returns a list of members containing the given annotation.
+     *
+     * @param annotation find members containing this annotation
+     */
+    private fun getMembersWithAnnotation(annotation: Class<out Annotation>):
```
ivanporty: Could we use `KClass` instead since we only check Kotlin annotations here?

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
```diff
@@ -64,33 +72,80 @@ class ContainerToolsRule(private val testInstance: Any) : TestRule {
...
+        List<KProperty1<out Any, Any?>> =
+        testInstance::class.declaredMemberProperties.filter { member ->
+            member.annotations.filter {
+                annotation.isAssignableFrom(it::class.java)
```
ivanporty: Could we use the style used previously, with `it` instead of assignable from and Java class - i.e. in the manner of `(member.annotations.filter { it is annotation }`?

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
```diff
@@ -64,33 +72,80 @@ class ContainerToolsRule(private val testInstance: Any) : TestRule {
...
+                member !is KMutableProperty<*>
+            ) {
+                throw IllegalArgumentException(
+                    "@TestFile can only annotate mutable fields of type " +
+                        "java.io.File"
+                )
+            }
+
+            val annotation: TestFile? = member.findAnnotation()
+            val directory: File = FileUtil.createTempDirectory(directoryName, null)
```
ivanporty: Will this directory be deleted on the test end?

ivanporty:
@etanshaul should we also request a review from our Kotlin readability contributors?

etanshaul:
Yeah. let me address your comments first then we can see if there is any Kotlin'y things that we think should be reviewed. thanks.

skaffold-editing/src/main/kotlin/com/google/container/tools/skaffold/editing/SkaffoldContextType.kt
etanshaul: good suggestion. done

skaffold-editing/src/main/resources/liveTemplates/Skaffold.xml
```diff
@@ -0,0 +1,9 @@
+<templateSet group="Skaffold">
```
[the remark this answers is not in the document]
etanshaul: i don't think its a hard rule. I mostly just followed convention. To see what this translates to concretely:
screen shot 2018-10-26 at 2 08 10 pm

The group name is `Skaffold` (human readable). what the template name (more of an ID) is `skaffold`.

This seems to be how most of the other templates do things.

skaffold-editing/src/main/kotlin/com/google/container/tools/skaffold/editing/SkaffoldLiveTemplatesProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+private const val SKAFFOLD_LIVE_TEMPLATE_PATH = "liveTemplates/Skaffold"
+
+/**
+ * A Skaffold [DefaultLiveTemplatesProvider] that provides the live templates used for generating /
+ * editing Skaffold configuration files.
+ */
+class SkaffoldLiveTemplatesProvider : DefaultLiveTemplatesProvider {
+
+    override fun getDefaultLiveTemplateFiles(): Array<String> {
+        return arrayOf(SKAFFOLD_LIVE_TEMPLATE_PATH)
```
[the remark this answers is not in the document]
etanshaul: done

skaffold-editing/src/main/kotlin/com/google/container/tools/skaffold/editing/SkaffoldLiveTemplatesProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    override fun getHiddenLiveTemplateFiles(): Array<String>? {
+        return emptyArray()
```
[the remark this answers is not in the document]
etanshaul: done

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
etanshaul: yes. that is managed by the platform

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
etanshaul: so what you are suggesting is actually different behavior.

`annotation.isAssignableFrom(it::class.java)`

that is checking that the current member is a subclass of the passed in `annotation`. The passed in annotation could be `TestService`, or `TestFile`, for example. This allows this method to be reusable.

Does that clear this up or do you still see issues with it?

ivanporty: It works, just a bit confusing why we need to support subclassing for annotations, what if we just check for straight class match (i.e. `is` as `instanceof` for Kotlin), won't this be enough for us?

etanshaul: hmm, I see what you're saying now. i think you're right. Let me update.

etanshaul: aha. now I remember. I tried that myself initially, but the problem is that this code snippet:

`(member.annotations.filter { it is annotation }`

is not allowed. the part after the `is` needs to be a type, not a variable. So as far as I could tell what I have is the only way to do it. If you can think of a better way I'll welcome it.

ivanporty: Yeah, it looks like a logical case for a class, but seems not to be working. I'm ok with it in the current form then, maybe we can consult Kotlin reviewers on that as well.

etanshaul: I updated it to use `#isInstance` instead of `#isSuperclassOf` to hopefully make it less confusing.

Regarding using a class variable in the `x is y` expression, I think its the same limitation that java imposes on `instanceof`. If you need to use a class instance then you need to do `clazz.instanceOf(foo)`.

balopat:
> @balopat
> 
> This PR creates a bare bones `skaffold.yaml` file for the user to get started via IntelliJ's [Live Template](https://www.jetbrains.com/help/idea/using-live-templates.html) mechanism (see the PR description for more details).
> 
> Currently I have it creating the following:
> 
> 
> I was wondering if you could comment on what a good getting started `skaffold.yaml` should look like based on best practices. For example, should it default to using GCB?

The above looks great, that is the minimum possible config. No, we shouldn't default to GCB, I think local docker makes more sense to default to.

PR #56, opened 2018-10-29
Skaffold command line state and process start up

### LINKED ISSUE #3: Create generic Skaffold run configuration suitable for all JB IDEs
High level features:
1. Available from drop-down list of run/debug configurations
2. Exposes basic configuration UI (conceivably with skaffold.yaml location, environment variables)
3. Compatible with maximum number of IDEs (possibly with all of them)
4. Extensible to support language specific features (build image with JIB for Java etc.)

ivanporty:
Closes #3, #12, #13.
This PR completes initial set of work for supporting Skaffold run/deploy from within all Jetbrains IDEs.

Introduces `SkaffoldCommandLineState` class which uses built-in `CommandLineState` base class functionality to provide command line execution support, standard console window and graceful stop of the resulting Skaffold process. The class implementation converts Skaffold run configuration into running Skaffold process.

A few changes made to `ContainerToolsRule` in order to support complete mocks as replaced services - this needs extraction of actual class name since Mockk framework appends random `ByteBuddy` suffixes to class names.

`AbstractSkaffoldRunConfiguration` now explicitly calls `XmlSerializer` since base class doesn't persist settings by itself.


skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -76,6 +79,18 @@ abstract class AbstractSkaffoldRunConfiguration(
...
+    override fun readExternal(element: Element) {
+        super.readExternal(element)
+
+        XmlSerializer.deserializeInto(this, element)
+    }
+
+    override fun writeExternal(element: Element) {
+        super.writeExternal(element)
+
+        XmlSerializer.serializeInto(this, element)
```
etanshaul: Can you use a `PersistantStateComponent` instead?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineStateTest.kt
```diff
@@ -0,0 +1,135 @@
...
+
+        skaffoldCommandLineState = SkaffoldCommandLineState(
+            mockExecutionEnvironment,
+            SkaffoldExecutorSettings.ExecutionMode.DEV
+        )
+
+        val exception = expectThrows(
+            ExecutionException::class,
+            ThrowableRunnable { skaffoldCommandLineState.startProcess() })
+        assertThat(exception.message).isEqualTo(message("skaffold.unsupported.run.settings"))
```
etanshaul: instead of using the message from the bundle in the unit test, I think its better to have the message string again in the unit test (this is the pattern we've used in the other plugin too). This way, we also are testing that the message in the bundle matches what we expect.

You should also then be able to get rid of `@VisibleForTesting` on `#message`

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
```diff
@@ -0,0 +1,76 @@
...
+) : CommandLineState(environment) {
+    public override fun startProcess(): ProcessHandler {
+        val runConfiguration = environment.runnerAndConfigurationSettings?.configuration
+        val projectBaseDir = environment.project.baseDir
+        // ensure the configuration is valid for execution - settings are of supported type,
+        // project is valid and Skaffold file is present.
+        if (runConfiguration == null || runConfiguration !is AbstractSkaffoldRunConfiguration ||
+            projectBaseDir == null
+        ) {
+            throw ExecutionException(message("skaffold.unsupported.run.settings"))
```
etanshaul: under what circumstances would this message be shown to the user? is it possible to be more specific with the message? As is, I don't know if I would know how to start resolving it if I saw this message.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
```diff
@@ -0,0 +1,76 @@
...
+            configFile, projectBaseDir
+        )
+
+        SkaffoldExecutorService.instance.executeSkaffold(
+            SkaffoldExecutorSettings(
+                executionMode,
+                skaffoldConfigurationFilePath,
+                workingDirectory
+            )
+        ).let {
```
etanshaul: maybe a question for Kotlin readability people, but this (perhaps due to my Kotlin ignorance) looks less readable than just assigning the return to a val and then returning a `KillableProcessHandler` after. wdyt?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.intellij.execution.configurations.CommandLineState
+import com.intellij.execution.process.KillableProcessHandler
+import com.intellij.execution.process.ProcessHandler
+import com.intellij.execution.runners.ExecutionEnvironment
+import com.intellij.openapi.vfs.LocalFileSystem
+import com.intellij.openapi.vfs.VfsUtilCore
+import java.io.File
+
+/**
+ * Generic [CommandLineState] implementation for running Skaffold on command line based on the given
```
etanshaul: > Generic [CommandLineState] implementation for running Skaffold on command line ..

This sounds a bit confusing to me since I first thought you meant that this implementation is generic (not that `CommandLineState` is generic.

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
```diff
@@ -151,13 +151,23 @@ class ContainerToolsRule(private val testInstance: Any) : TestRule {
...
+     * In case of using [mockk] for mocking, tries to extract original class name from a randomly
+     * assigned mocked class name.
+     *
      * @param newInstance the new instance to register
      */
     private fun setService(newInstance: Any) {
         with(ApplicationManager.getApplication().picoContainer as MutablePicoContainer) {
-            unregisterComponent(newInstance::class.java.name)
+            var javaClassName = newInstance::class.java.name
+            // when using Mockk, the class name for mocks is random and uses ByteBuddy suffix.
```
etanshaul: interesting. I wonder why it was working previously for `@TestServices`s?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineStateTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        // Skaffold executor answer mocks
+        val mockSkaffoldProcess: SkaffoldProcess = mockk(relaxed = true)
+        every {
+            mockSkaffoldExecutorService.executeSkaffold(capture(skaffoldSettingsCapturingSlot))
+        } answers { mockSkaffoldProcess }
+    }
+
+    @Test
+    fun `unsupported run configuration type throws execution exception`() {
+        every { mockRunnerSettings.configuration } answers { mockk<RunConfigurationBase>() }
```
etanshaul: > mockk<RunConfigurationBase>()

Is that return what is causing the exception to be thrown in `startProcess` (presumably because its not an `AbstractSkaffoldRunConfiguration`?)? I wonder if it would be cleaner to declare the mock at the class level with a nice name to help readability? Something like `nonSkaffoldRunConfiguration`

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
etanshaul: also, what do you mean by the the name being random? presumably the prefix before the byteBuddy junk is always equal to the original classname?

etanshaul: oh I also just had this thought (sorry for the barrage of comments):

Presumably mockito also had this problem with whatever bytecode manipulation framework it uses (cglib or whatever). I recall that mockito mocks' classname has a suffix as well (something like `...EnhancedByMockito...`). So I guess I'm wondering why this needs treatment that we didn't need in the Cloud Tools plugin?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
ivanporty: Sure, let me rephrase that.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
ivanporty: I agree, this constructs are nice but they hide quite a lot. Let's revert to a more traditional style here.

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineStateTest.kt
ivanporty: Sounds good, I thought about that too, and probably we won't be changing the messages a lot, so this shouldn't be a maintenance issue.

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineStateTest.kt
ivanporty: You mean define it as a non-Skaffold configuration as a separate class, or inner class?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
ivanporty: This is going to be shown when you click `Run` and something is very wrong, i.e. mismatch in configuration type or empty project directory etc. Very unlikely to happen, probably in case of some bad migration or issue. What about `Skaffold run configuration is invalid. Please edit to fix errors, or re-create new one.`?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineStateTest.kt
etanshaul: Maybe I misunderstood what the code is doing:

It looks like you are returning a mock when `mockRunnerSettings.configuration` is called (the equivalent of mockito's `when(foo).thenReturn(mock(bar))`? I was suggesting that instead of inline mocking, do the mocking as an annotation at the top with a nicely named variable.

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
ivanporty: So the name of the mockk class is formed to something like this: `com.google.container.tools.skaffold.SkaffoldExecutorService$ByteBuddy$R5YLewgP$ByteBuddy$7xGX0mYJ`. Original class name is there, but it's appended with this ByteBuddy suffixes. Possibly Mockito has a different approach to this.

It only worked with `Spyk`, we haven't tried true `Mockk` with `TestService` before, so this came out.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
etanshaul: I wonder if its so rare/bad of a situation that it warns an IDE exception? If not, we could say something along the lines of `Your Skaffold run configuration is corrupted` followed by some wording to recreate a new one like you suggested.

Not to self: we need to add the plumbing for IDE exception reporting like with have in CT4IJ. I'll create an issue.

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
etanshaul: https://stackoverflow.com/questions/46191265/how-to-get-original-class-name-from-a-mocking-object 

That suggests that mockito does something similar. In any case, its not important if you've verified that it indeed needs this fix.

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineStateTest.kt
ivanporty: ah, so you meant variable with a cleaner name to make sure it's obvious this is a source of the issues? Sounds good, let me do that.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
ivanporty: Hard to say really, it looks like it could only happen in case of some deep state corruption. So you suggest we throw some sort of `RuntimeException` here to invoke IDE reporting mechanics, correct?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineStateTest.kt
etanshaul: yep. also in there is the suggestion to stick to annotation based mocking exclusively. I'm not sure I have a good reasoning as to why (so I could be misguided here), but its just a carry over from a rule I followed using mockito.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
etanshaul: I think I'm ok in this situation just notifying the user that their state is corrupted and to re-create the run configuration.

(btw if we had gone the route of triggering an IDE error, `log.error()` would have worked too)

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineStateTest.kt
ivanporty: I can extract mock Skaffold dev configuration, it used in 2 tests, but this one is very local. I'd say they it would make it hard to read to put all possible one use mocks on top of the class, wdyt? It seems like a good name and a local instantiation can make it easier to follow the logic here.

etanshaul: I'd say go with what you prefer since it ultimately isn't a huge deal.

Personally I prefer declaring all the mocks at the class level. It makes it easy to quickly see what is being mocked in the test class. I also find the mixing of two mocking styles a bit messy.  But I also understand your point about it being local..

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
ivanporty: Actually, I double checked what we did in Cloud Tools, and we used the original type of the field instead of the mocked object type. It should work here as well, I'll do the change.

build.gradle.kts
```diff
@@ -38,6 +38,7 @@ allprojects {
         type = project.properties["ideaEdition"].toString()
         version = project.properties["ideaVersion"].toString()
         intellijRepo = project.properties["intellijRepoUrl"].toString()
+        alternativeIdePath = "/usr/local/google/home/ivanporty/Documents/GoLand-2018.1.5/"
```
etanshaul: probably don't want to commit this :)

ivanporty: lol, true

skaffold/src/main/resources/messages/SkaffoldBundle.properties
```diff
@@ -24,3 +24,4 @@ skaffold.configuration.label=Skaffold configuration:
 
 skaffold.no.file.selected.error=Skaffold configuration file is not selected.
 skaffold.invalid.file.error=Skaffold configuration file doesn't exist or invalid.
+skaffold.corrupted.run.settings=Your Skaffold run configuration is corrupted. Please edit or re-create it to fix.
```
etanshaul: I wonder if editing it would do much to help? If not, then maybe just suggest to re-create it.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
```diff
@@ -0,0 +1,76 @@
...
+            throw ExecutionException(message("skaffold.corrupted.run.settings"))
+        }
+        if (runConfiguration.skaffoldConfigurationFilePath == null) {
+            throw ExecutionException(message("skaffold.no.file.selected.error"))
+        }
+
+        val workingDirectory = File(environment.project.basePath)
+        val configFile = LocalFileSystem.getInstance()
+            .findFileByPath(runConfiguration.skaffoldConfigurationFilePath!!)
+        // use project dir relative location for cleaner command line representation
```
etanshaul: > for cleaner command line representation

You mean as opposed to an absolute path? Does Skaffold allow both relative an absolute paths? If so, sgtm.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
ivanporty: We could, but it would essentially do the same thing, and there is a dangerous mix of `loadState()` methods from `RunConfigurationBase` and `PersistantStateComponent`. The only difference is that we would probably create the DOM Element ourselves.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
ivanporty: It does, yes.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
```diff
@@ -0,0 +1,76 @@
...
+ * @param environment Execution environment provided by IDE
+ * @param executionMode Skaffold  mode execution, i.e. DEV
+ */
+class SkaffoldCommandLineState(
+    environment: ExecutionEnvironment,
+    val executionMode: SkaffoldExecutorSettings.ExecutionMode
+) : CommandLineState(environment) {
+    public override fun startProcess(): ProcessHandler {
+        val runConfiguration = environment.runnerAndConfigurationSettings?.configuration
+        val projectBaseDir = environment.project.baseDir
```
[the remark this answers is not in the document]
ivanporty: You're right, IDEA shows them for you and it almost looks like they are there. Let me add them, since there is a heavy mix of virtual files, just files and strings here indeed.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
etanshaul: ok. lgtm.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -76,6 +79,18 @@ abstract class AbstractSkaffoldRunConfiguration(
      * Persisted Skaffold config file absolute path for Skaffold run configurations.
      * See more at [com.intellij.openapi.vfs.VirtualFile.getPath]
      */
-    @Attribute("skaffoldConfigurationFilePath")
+    @get:OptionTag("skaffoldConfigurationFilePath")
```
etanshaul: just for my own knowledge. what is this annotation?

ivanporty: I think this is called Annotation Use-site Targets per Kotlin docs, they are using this for persisting Kotlin based properties. I think this limits this annotation to certain targets generated for annotation, for this case for getters, I saw this in `RunConfigurationOptions`.

etanshaul: I actually meant the OptionTag part. The other info is good to know though!

etanshaul: https://github.com/JetBrains/intellij-community/tree/master/platform/util/src/com/intellij/util/xmlb/annotations#readme

Seems like this is for backwards compatibility. But I guess we have no choice because we are stuck with the base class that is using this style?

ivanporty: Hmm, you're right, apparently it just works without this tags at all, and in completely similar manner. Thanks for the hint, this makes it easier!

etanshaul: thats strange. I wonder why it works without it? how does it know what to serialize now?

etanshaul: sg

PR #59, opened 2018-10-30
Expand version range to include upcoming 2018.3 IDE, allow older IDEs to use the plugin.

ivanporty:
IntelliJ Gradle plugin was enforcing minimum version of the IDE used to build it (2018.2+) to run the plugin if `updateSinceUntilBuild` property is not set to `false`.


ivanporty:
Yep, but already in beta, so should be in soon. Thanks!

Issue #25, opened 2018-09-25
Create and use Skaffold action icons.


ivanporty:
The following icons, aside from the original Skaffold icon (made from logo) are needed:
* Skaffold continuous run image, separating it from single run configuration
* Skaffold action icon for manual trigger to re-deploy

ivanporty:
Deprecated. https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/2369

PR #63, opened 2018-10-31
Better separate dev and deploy mode with icons, title in editor

### LINKED ISSUE #57: Both Skaffold run configuration types look the same
I created a new `skaffold dev` type run configuration and gave it some name, something like `my Skaffold config`.

A few moments later I went in to edit it and I couldn't remember if it was a `dev` or `run` type.

Perhaps we should have a label in the UI to make this clear

ivanporty:
Work towards #25, fixes #57.

Slightly different icon for DEV mode (with a green light), and added a description text in the settings editor to see what exactly it is. We can probably add a bit of explanatory text too.

skaffold-dev-icon-text


skaffold/src/main/resources/messages/SkaffoldBundle.properties
```diff
@@ -17,8 +17,8 @@
 skaffold.run.config.general.name=Skaffold
 skaffold.run.config.general.description=Runs Skaffold
 
-skaffold.run.config.single.run.name=Skaffold Single Run
-skaffold.run.config.dev.run.name=Skaffold Continuous
+skaffold.run.config.single.run.name=Skaffold Single Deployment
```
etanshaul: How about just "Skaffold Deployment"

etanshaul:
I'm not 100% sure about the icon thoguh. I like the idea of making it visually distinct, but that implementation makes it a bit confusing for me. The green light doesn't make me think of it being local dev (and why not have the it on skaffold run instead etc.).

I'm don't really have an alternative suggestion though yet. I'm wondering if just having the different default text is enough?

Can you remind me why we decided not to just have a single run config type with a drop-down (or radio button etc.) to select 'dev' vs 'run'? For some reason I keep going back to that when trying to come up with a good solution.

skaffold/src/main/resources/messages/SkaffoldBundle.properties
etanshaul: or even more precisely: `Deploy to Kubernetes with Skaffold`? Seems like we have the real estate.

ivanporty: Good idea, we probably should go straight to Kubernetes.
How about selection is "Kubernetes with Skaffold" -> opens up to two configurations "Deploy to Kubernetes with Skaffold" and "Develop on Kubernetes with Skaffold"?

etanshaul:
Ah yeah. Regarding the 2 separate configs that makes a lot of sense. Let's definitely keep that for now.

Are you ok putting the icon effort to the side for now until we can have some UX feedback? I am also hesitant about a k8s icon for deploy because it is also using Skaffold entirely so it still feels a bit arbitrary.

If you are ok with that, we can focus this PR on just adding some nice labeling to the run config.

etanshaul:
I'd also almost be willing to consider using the k8s icon instead of skaffold icon (for both), but that's a separate discussion.

ivanporty:
Sounds good, let me rephrase with Kubernetes included and update the screenshot.

ivanporty:
The result:

skaffold-run-helper-text

ivanporty:
Also needed to add separate IDs for both `run` and `dev` configurations, without them, any change in the run configuration name led to previous versions being invalid.

etanshaul:
nice!

Small nit with `Develop continuously with ...` -> `Continuous development with ...`? 

Thats also the wording they use on the Skaffold github readme. "Continuous X" seems to be a more common phrasing.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldUiResources.kt
```diff
@@ -23,6 +23,7 @@ import org.jetbrains.annotations.PropertyKey
 import java.util.ResourceBundle
 
 internal val SKAFFOLD_ICON = IconLoader.getIcon("/icons/skaffold.png")
+internal val SKAFFOLD_DEV_ICON = IconLoader.getIcon("/icons/skaffold-dev.png")
```
etanshaul: can this be removed for now?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurationType.kt
```diff
@@ -53,19 +54,29 @@ class SkaffoldRunConfigurationType : ConfigurationType {
  * at [SkaffoldSingleRunConfiguration].
  */
 class SkaffoldSingleRunConfigurationFactory(type: ConfigurationType) : ConfigurationFactory(type) {
+    val RUN_ID = "google-container-tools-skaffold-run-config-single"
```
etanshaul: nit: "...-run" instead of "...-single"?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldUiResources.kt
ivanporty: done.

PR #79, opened 2018-12-05
Detects presence of Skaffold configuration file(s)

### LINKED ISSUE #7: Detect presence of Skaffold configuration file(s) and ask to create Skaffold run configuration
If project is opened or new files appear, detect if Skaffold configuration YAML is present, and prompt a user to create Skaffold run configuration, using pop-up notification.

ivanporty:
Fixes #7.

Detects presence of Skaffold configuration file(s) and uses notification on project open to ask to create Skaffold run configurations, either for both development and deployment, or one of them.

Example: project with existing Skaffold configuration but not prior Skaffold run configurations opened, the following notification shows:

skaffold-detect-prompt

User has an option to create both or one of the configuration, development target being the most visible (at least in the current 2018.3 implementation of Notification). If both configurations are selected, both are created, development target becomes selected in run configurations combobox:

skaffold-auto-configs

Follow up PR: detect presence of a new Skaffold file when it's created in an already existing project.


etanshaul:
Nice! Should the default action be to create both of them I wonder?

Also, I wonder if perhaps we should align the wording more with the default "Framework Detected" dialog:

image

After clicking configure in the above, you then select the appengine-web.xml file you wish to configure:
image

Can you remind me too why don't want to use the framework support mechanism to achieve this? "Skaffold" framework support absolutely shouldn't be required to have access to all the plugin features, but I wonder if we can use it just for the purposes of detecting the skaffold config?

etanshaul:
just gave it for a spin too. works great!

ivanporty:
Regarding frameworks, I thought about using framework detector to detect for files, but this indeed requires a framework. The framework idea as it's implemented in IDEA might look confusing to non-Java developers though since we target all JetBrains IDEs. Even more concerning, the support for "frameworks" and "facets" might not be complete for other IDEs, while simple run configurations are 100% supported across all IDEs. This is something we can discuss with JetBrains team on their favorite approach. 

As an example of something working across all IDEs, I think Docker plugin does not add any "Docker" frameworks, although it detects files and adds run configurations.

ivanporty:
+1 on making both configs the most visible option (again, highly version/implementation specific.)

etanshaul:
ah! thats the reason...the other IDEs. I wonder what even happens in other IDEs (like GoLand for instance). Is framework support just not supported?

ivanporty:
I didn't see anything "framework" in GoLand, there might be something in PyCharm (Ultimate), but this needs a lot of research. Unless there's a strongly recommended approach from JetBrains (we can ask them?), I think non-framework approach seems to be cleaner and simpler from both conceptual and implementation point of view.

etanshaul:
agree 100%. if its only supported in IDEA then we shouldn't use (keeping it simple with a single language/ide-agnostic approach).

etanshaul:
How about something like this:

```
Skaffold configuration detected.

Use Skaffold do deploy your applications to Kubernetes. Skaffold can also assist with rapid development by watching the dependencies of your docker image for changes, and redeploying your application to a Kuberenetes cluster for testing.
```

Then the actions would be as you have them, except that the second like will just say `configure Skaffold` and will create both run configs.

wdyt?

etanshaul:
Also we could consider simplifying the action by getting rid of the actions drop down (can we do this?) and just having the "configure Skaffold" action (as described in the previous comment). The user can then decide later if they only need one of them (and delete it manually). This way there are less things to think about up front. Just and idea though..

ivanporty:
It looks like the most notifications in IDEA/other IDEs take the shorter text approach - otherwise, except for the first sentence, the text will always appear clipped and not quite pleasant to see unless you close it and then re-read from event log. I can make a screenshot with a demo of that.

As for only suggesting one Skaffold action that creates both actions, I think this is a good idea and leaves user with less decision making. However, what if we still keep Kubernetes wording matching what we have in configurations naming? I.e. `Create run configurations for Kubernetes with Skaffold `

etanshaul:
Yeah. clipping would be no good. I like the wording you suggested (I assume you mean that would be the action name?). If so, maybe the "create configurations?" part of the body of the notifications is redundant since the action is saying clearly that it will create the configurations.

ivanporty:
Proposal:

skaffold-prompt-2

etanshaul:
I like it. Although it would be nice if it said something about the 'skaffold dev' feature. Can the description have bullets?


```
Skaffold configuration detected. Use Skaffold to:
- Deploy your applications to Kubernetes
- Update your deployed application continually during development
```

Second sentence wording adapted from the Skaffold github readme.

Is this too long?

ivanporty:
It looks ok, truncated but can easily be opened with an arrow, i.e. initial state:
skaffold-detect-ellipsis


opened text state:

skaffold-detect-list

List bullets rendering is not ideal (look truncated) but looks like minor thing. WDYT?

etanshaul:
I like it actually. We can tweak it later based on feedback, if you are ok with it.

etanshaul:
nit: deploy in the first bullet -> Deploy

ivanporty:
Ok, I'm good with that text too, let's go with that for now.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldConfigurationDetector.kt
```diff
@@ -0,0 +1,157 @@
...
+        val runnerAndConfigSettings = RunnerAndConfigurationSettingsImpl(
+            RunManagerImpl.getInstanceImpl(project),
+            runConfiguration
+        )
+        getRunManager(project).addConfiguration(runnerAndConfigSettings)
+        getRunManager(project).selectedConfiguration = runnerAndConfigSettings
+    }
+
+    @VisibleForTesting
+    fun createNotification(title: String, message: String): Notification {
```
etanshaul: nit: function expression (`fun ..(): Notificatio =`)

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldConfigurationDetector.kt
```diff
@@ -0,0 +1,157 @@
...
+        return NOTIFICATION_GROUP.createNotification(
+            title, null /* subtitle */, message, NotificationType.INFORMATION
+        )
+    }
+
+    @VisibleForTesting
+    fun getRunManager(project: Project): RunManager {
```
etanshaul: same comment. simple function expression

etanshaul:
looks great. just a couple minor comments, update to the wording based on the discussion, then good to go

ivanporty:
thanks, done.

PR #84, opened 2018-12-07
Adds tail logs support for single deployment mode

### LINKED ISSUE #69: Add tail logs support for single deployment mode
Single run (deployment) mode does not stream logs by default, which can be useful to assess initial progress of the deployment. Stopping the process will then stop the logs streaming, but leave the original deployment intact. We can add a simple option for single run mode to enable log tailing.

ivanporty:
Fixes #69.

Adds UI option to enable log tailing for single (deployment) mode:

skaffold-tail-logs-ui

Once deployment (successfully) completes, logs are streamed until process is killed. Deployment itself stays working:

skaffold-tail-console

### DIFF
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
@@ -64,8 +64,8 @@ abstract class SkaffoldExecutorService {
                     }
             }
 
-            settings.tailLogsAfterDeploy?.let {
-                if (it == true) add("--tail")
+            settings.tailLogsAfterDeploy?.let { tailLogs ->
+                if (tailLogs) add("--tail")
             }
         }
 
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
@@ -63,6 +63,10 @@ abstract class SkaffoldExecutorService {
                         add("${label.key}=${label.value}")
                     }
             }
+
+            settings.tailLogsAfterDeploy?.let {
+                if (it == true) add("--tail")
+            }
         }
 
         return SkaffoldProcess(
@@ -97,7 +101,8 @@ data class SkaffoldExecutorSettings(
     val executionMode: ExecutionMode,
     val skaffoldConfigurationFilePath: String? = null,
     val workingDirectory: File? = null,
-    val skaffoldLabels: SkaffoldLabels? = null
+    val skaffoldLabels: SkaffoldLabels? = null,
+    val tailLogsAfterDeploy: Boolean? = null
 ) {
 
     /** Execution mode for Skaffold, single run, continuous development, etc. */
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
@@ -47,9 +47,9 @@ open class BaseSkaffoldSettingsEditor<C : AbstractSkaffoldRunConfiguration>(
     @VisibleForTesting
     val skaffoldFilesComboBox = SkaffoldFilesComboBox()
 
-    lateinit var basePanel: JPanel
+    protected lateinit var basePanel: JPanel
 
-    val extensionComponents: MutableMap<String, JComponent> = mutableMapOf()
+    private val extensionComponents: MutableMap<String, JComponent> = mutableMapOf()
 
     /**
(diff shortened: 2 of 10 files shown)


skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -39,6 +39,9 @@ class SkaffoldSingleRunConfiguration(
     name: String
 ) : AbstractSkaffoldRunConfiguration(project, factory, name) {
 
+    /** Single run tail logs option to stream logs after deployment until stopped. */
+    var tailDeploymentLogs: Boolean = false
```
etanshaul: does this mean that it defaults to unchecked? wdyt about defaulting it to tail? also maybe add a unit test to confirm the default value (whatever we decide on).

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
```diff
@@ -37,27 +38,46 @@ import javax.swing.JComponent
...
+            }
+        }
+
         basePanel.border = IdeaTitledBorder(editorTitle, 0, Insets(0, 0, 0, 0))
 
         return basePanel
     }
 
-    override fun applyEditorTo(runConfig: AbstractSkaffoldRunConfiguration) {
+    override fun applyEditorTo(runConfig: C) {
```
etanshaul: why type `C`?

etanshaul:
this is awesome. the `--tail` option really simplifies things

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
ivanporty: `C[onfigurationType]`, to simplify child editors. You mean `C` is confusing instead of typical `T`?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
ivanporty: Sure, good idea on unit test.
As for the default value, this is derived from Skaffold approach - they do not tail by default for `run`, probably makes sense for us to keep some degree of familiarity for existing Skaffold users?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
etanshaul: Ah its a generic type. I think T would be more standard, but I don't feel strongly

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
etanshaul: sure. sgtm. We can always change it later if it turns out to be more useful.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -63,6 +63,10 @@ abstract class SkaffoldExecutorService {
                         add("${label.key}=${label.value}")
                     }
             }
+
+            settings.tailLogsAfterDeploy?.let {
+                if (it == true) add("--tail")
```
etanshaul: you can't just do `if (it) ...`?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/BaseSkaffoldSettingsEditor.kt
ivanporty: According to Google Java style (http://google.github.io/styleguide/javaguide.html#s5-naming), naming either allows single letter (no strict recommendation on which exact letter though), or type name followed by a `T`, i.e. `RunConfigurationT`. I think both are fine, second being a bit more clear, but I think most of the developers already got used to `T`, `E` etc. We can still switch to `T` for maximum compatibility.

etanshaul: ah. ok cool. Yeah Ive never seen the `RunConfigurationT` style before - I don't really like it either myself.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
etanshaul: +1. your suggested snipped is a nice improvement

PR #85, opened 2018-12-10
Set up analytics infrastructure

etanshaul:
progress towards #47 

Initial setup of analytics infrastructure. Consumes shared analytics library (currently bundled directly as a jar).

TODO (next PR):
- Set up UI for opt-in/out
- Consume opt-in/out setting
- Ensure analytics inter operate with Cloud Tools for IntelliJ plugin


core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
```diff
@@ -0,0 +1,49 @@
...
+    private val properties: Properties = Properties()
+
+    companion object {
+        private const val DEFAULT_PROPERTIES_FILE_NAME = "/config.properties"
+    }
+
+    init {
+        this.javaClass.getResourceAsStream(propertyFilePath)?.use {
+            properties.load(it)
+        } ?: throw IllegalArgumentException(
```
ivanporty: Could we use `log.error` or just throwing an exception will show the error level message anyway? Not sure which one looks like a better fit in here.

core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerManagerService.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.google.common.annotations.VisibleForTesting
+import com.google.container.tools.core.properties.PropertiesFileFlagReader
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.components.ServiceManager
+
+/**
+ * Application service that manages the status of usage tracking in the plugin.
+ */
+class UsageTrackerManagerService(
+    private val propertyReader: PropertiesFileFlagReader = PropertiesFileFlagReader()
```
ivanporty: This looks a bit heavy for a single value dependency - could we use `analyticsId` as a parameter directly instead?

ivanporty: And, conceivably, just expose `PropertiesFileFlagReader` as a simple service?

core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.Properties
+
+/**
+ * Reads property values from the specified properties file.
+ *
+ * @param propertyFilePath the path to the properties file; defaults to 'config.properties' located
+ * in the root of the module's resources
+ */
+class PropertiesFileFlagReader(
```
ivanporty: I feel this name is a bit confusing - it mentions `flags`, but all javadocs just say "property". Property looks to be very familiar concept, what if we just use property and maybe also hint this is out plugin properties, not some general purpose class - i.e. `ContainerToolsProperties` or something similar?

core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
```diff
@@ -0,0 +1,49 @@
...
+            properties.load(it)
+        } ?: throw IllegalArgumentException(
+            "Failed to load plugin property configuration file: $propertyFilePath"
+        )
+    }
+
+    /**
+     * Return the property value given the passed in name.
+     */
+    fun getFlagString(propertyName: String): String? = properties.getProperty(propertyName)
```
ivanporty: ditto about flags / vs. documented as properties.

core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.intellij.openapi.application.PermanentInstallationID
+import com.intellij.openapi.components.ServiceManager
+
+/**
+ * Application service that initializes the [UsageTracker] for analytics. Sets up core components
+ * of analytics collection such as the analytics ID, user agent, client ID etc.
+ */
+class UsageTrackerProvider {
+
+    val usageTracker: UsageTracker
```
ivanporty: This property will need to be accessed to send the actual data, correct?

core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+            .pluginVersion(PluginInfo.instance.pluginVersion)
+            .clientId(PermanentInstallationID.get())
+            .userAgent(PluginInfo.PLUGIN_USER_AGENT)
+            .build()
+
+        usageTracker = UsageTracker.create(settings)
+    }
+
+    companion object {
+        private const val PAGE_HOST = "virtual.intellij"
```
ivanporty: We don't expect this to change as far as I remember, so no need to place this in configuration file I assume?

etanshaul: right I don't anticipate this ever changing.

core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
etanshaul: Completely agree. How about just `PropertiesFileReader`?

core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
etanshaul: yeah. how about `getPropertyValue`, and the param is called `propertyKey`?

core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerManagerService.kt
etanshaul: Could you clarify what part of this is heavy? Do you mean the UsageTrackerManagerService class in general?

core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
etanshaul: Yeah. we could presumably do both too (I'll think about it).

But this reminds me of something else. Currently, the way the skaffold process execution works, no exceptions are every bubbled up to the IDE. Instead, when you try to run the process, it shows this little red notification thing (and in this case where I am throwing an exception, the notification will show the exception message). We may want to think about possibly letting certain exceptions bubble through so that we get notified, in the case of truly unanticipated behavior (like failure to load our property file).

core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
ivanporty: sounds great!

core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerManagerService.kt
ivanporty: I mean we pass a whole property file reader to this object just to extract a single string value, and it seems we don't anticipate many new properties or access to a reader itself in this class which is why it's a bit confusing to see we pass a reader.

How about having plugin properties as a service or just a singleton companion object, and then just pass in actual analytics id value as a parameter?

core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
ivanporty: Sounds good! In case we only use this class as a source of properties for our plugin and don't anticipate to use as as general property reader, we can probably further customize its name as a plugin propeties something?

core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
ivanporty: Agree, good idea. For now as we run system default Skaffold, it's hard to justify IDE level error for most of the run errors - but we definitely should do that once we have more functionality and managed tools and can know more about unexpected errors. We should also think about error notifications too along with console red error messages.

core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerManagerService.kt
etanshaul: Ah. Yeah I have it passed in because I wanted to mock it for testing. Changing it to a service would work.

Wrt passing the analytics id value in. I think that will be fine. I'll try and refactor.

etanshaul:
@ivanporty I think I've addressed your comments

core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerProvider.kt
ivanporty: I think the example looks good. Probably we can add it here as a kotlin doc too?

etanshaul: done

core/src/main/kotlin/com/google/container/tools/core/properties/PropertiesFileFlagReader.kt
etanshaul: Right. In this case however, I think we'd still want to know since its not related directly to the skaffold run (its a side effect of running an analytics ping with an exception loading the properties file). I'll open an issue to track.

PR #86, opened 2018-12-10
Refactor API usage for automatically adding run configurations in 2018.2

ivanporty:
#7 enabled automatic addition of Skaffold configurations but used new API not available on 2018.2 family of IDEs. There is a more standard way to create configurations using factory IDs and configuration type which is used now and this enabled automatic configurations on 2018.2 as well as 2018.3.

### DIFF
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldConfigurationDetector.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldConfigurationDetector.kt
@@ -24,6 +24,7 @@ import com.google.container.tools.skaffold.run.SkaffoldRunConfigurationType
 import com.google.container.tools.skaffold.run.SkaffoldSingleRunConfigurationFactory
 import com.intellij.execution.RunManager
 import com.intellij.execution.configurations.ConfigurationFactory
+import com.intellij.execution.configurations.ConfigurationTypeUtil
 import com.intellij.execution.configurations.RunConfiguration
 import com.intellij.notification.Notification
 import com.intellij.notification.NotificationDisplayType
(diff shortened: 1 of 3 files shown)


skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldConfigurationDetector.kt
```diff
@@ -95,42 +94,60 @@ class SkaffoldConfigurationDetector(val project: Project) : ProjectComponent {
...
-        )
+    private fun createRunConfigurationFromSettings(
+        factoryId: String,
+        name: String,
+        skaffoldFilePath: String
+    ) {
+
+        val factory = findConfigurationFactoryById(factoryId)
+        if (factory == null) {
+            Logger.getInstance(SkaffoldConfigurationDetector::class.java)
```
etanshaul: probably create a logger variable and just call `log.warn` here?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldConfigurationDetector.kt
```diff
@@ -95,42 +94,60 @@ class SkaffoldConfigurationDetector(val project: Project) : ProjectComponent {
...
+                .warn("Skaffold configuration factory not found, plugin install is corrupted.")
```
etanshaul: should we make this an error instead of a warn since its truly something we don't expect to ever happen?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurationType.kt
```diff
@@ -53,7 +55,9 @@ class SkaffoldRunConfigurationType : ConfigurationType {
  * at [SkaffoldSingleRunConfiguration].
  */
 class SkaffoldSingleRunConfigurationFactory(type: ConfigurationType) : ConfigurationFactory(type) {
-    val RUN_ID = "google-container-tools-skaffold-run-config-run"
+    companion object {
+        val RUN_ID = "google-container-tools-skaffold-run-config-run"
```
etanshaul: const?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurationType.kt
```diff
@@ -68,7 +72,9 @@ class SkaffoldSingleRunConfigurationFactory(type: ConfigurationType) : Configura
  * See template configuration at [SkaffoldDevConfiguration].
  */
 class SkaffoldDevConfigurationFactory(type: ConfigurationType) : ConfigurationFactory(type) {
-    val DEV_ID = "google-container-tools-skaffold-run-config-dev"
+    companion object {
+        val DEV_ID = "google-container-tools-skaffold-run-config-dev"
```
etanshaul: const?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldConfigurationDetector.kt
ivanporty: I think so, yes, that would indicate some completely unexpected corruption / misconfg.

ivanporty:
done.

Issue #87, opened 2018-12-11
Setup release scripts and gradle release configuration


etanshaul:
- bash script to be executed by kokoro release job
- gradle release plugin
- gradle publishPlugin task configuration (called by bash script)
- setup backend kokoro config (not part of this repo)

PR #91, opened 2018-12-11
Setup release script and gradle release / publish tasks

etanshaul:
progress towards #87 

Missing from #87 is the backend kokoro config which is done separately (not in this repo).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -99,4 +99,3 @@ release {
 }
 
 inline operator fun <T : Task> T.invoke(a: T.() -> Unit): T = apply(a)
-
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -14,11 +14,14 @@
  * limitations under the License.
  */
 
+import net.researchgate.release.GitAdapter
+import org.jetbrains.intellij.tasks.PublishTask
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     id("org.jetbrains.intellij") version "0.3.7"
     id("com.diffplug.gradle.spotless") version "3.14.0"
+    id("net.researchgate.release") version "2.7.0"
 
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -79,3 +82,20 @@ dependencies {
...
+}
+
+release {
+    tagTemplate = "v\${version}"
+
+    val git: GitAdapter.GitConfig = getProperty("git") as GitAdapter.GitConfig
+    git.requireBranch = "/^release_v\\d+.*$/"
+}
+
+inline operator fun <T : Task> T.invoke(a: T.() -> Unit): T = apply(a)
```
ivanporty: Interesting, we convert apply to direct call for Task?

build.gradle.kts
```diff
@@ -79,3 +82,20 @@ dependencies {
 
     compile("com.google.protobuf:protobuf-java:2.5.0")
 }
+
+val intellijRepoChannel: String by project
+val publishPlugin: PublishTask by tasks
+publishPlugin {
```
ivanporty: So as I understand, IJ release process uses `net.researchgate.release`?

build.gradle.kts
etanshaul: Yeah. Makes the task definitions a bit cleaner

build.gradle.kts
etanshaul: right. that is the gradle release plugin, called by `gradlew release`. This does some tagging, updates the version in `settings.gradle` to the next version, pushes a PR to github. 

To do the actual release, we need to run the Kokoro release job (not yet setup for this repo). This release job will invoke the release.sh script in the PR. (previously when we were using Jenkins we had the ability to configure Kokoro to detect a "release" triggered from the gradle release plugin and automatically call the release job. But it wasn't possible with Kokoro last I checked).

PR #92, opened 2018-12-12
 Use smart mode action to only detect Skaffold after indexing is done.

### LINKED ISSUE #89: Create Skaffold configuration prompt is not shown if opened project is being indexed first
If project opened has dependencies that need long indexing or IDE is opened for first time and needs indexing of a core SDK, Skaffold files detection does not work yet. It need to only be run after initial project indexing is done and files are available.

ivanporty:
Fixes #89.

The fix is to use "smart mode action" which is only called after all heavy tasks preventing files from being found or indexed are complete - this includes projects opened first time in IDE or IDE first start up time when they index the whole SDK or all dependencies.


skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldConfigurationDetector.kt
```diff
@@ -57,15 +58,19 @@ class SkaffoldConfigurationDetector(val project: Project) : ProjectComponent {
...
-        if (skaffoldFiles.isNotEmpty()) {
-            val skaffoldRunConfigList: List<RunConfiguration> =
-                getRunManager(project)
-                    .allConfigurationsList.filter { it is AbstractSkaffoldRunConfiguration }
-            if (skaffoldRunConfigList.isEmpty()) {
-                // existing Skaffold config files, but no skaffold configurations, prompt
-                showPromptForSkaffoldConfigurations(project, skaffoldFiles[0])
+        // on project open there may be a lot of indexing and other preparation and files still
+        // not available, wait for "smart" mode for configuration identification
+        DumbService.getInstance(project).runWhenSmart {
```
etanshaul: nice! I recall doing something similar for services that didn't need to wait for any indexing at all: you'd simply implement the `DumbAware' marker interface. I wonder if there is something like this for a  "smart" service that we could use from a ProjectComponent

ivanporty: You're right, `DumbAware` can do its action when indexing is on in background! I guess it's more useful for menu items or toolbar actions so a user can see what is still available.

etanshaul: Right. I was just wondering for this PR if there was a marker interface that ProjectComponent could implement (much in the same way that we do for `DumbAware` actions) instead of having to wrap the code in `runWhenSmart`. I guess not though - I didn't see anything when quickly looking.

ivanporty: Agree, I looked initially as well and didn't see anything looking useful yet.

PR #93, opened 2018-12-12
Update release script to pull keystore values directly

etanshaul:
explained nicely by @chanseokoh here: b/115336591

### DIFF
--- a/kokoro/unix/release.sh
+++ b/kokoro/unix/release.sh
@@ -30,12 +30,9 @@
 
 ## Release script that publishes our plugins to Github and Jetbrains plugin repository
 
-## Assert required environment variables
-
-[ -z "$ANALYTICS_ID" ] && \
-    echo "ERROR: Releasing requires the ANALYTICS_ID environment variable" && exit 1;
-[ -z "$GITHUB_TOKEN" ] && \
-    echo "ERROR: Releasing requires the GITHUB_TOKEN environment variable" && exit 1;
+## Export required environment variables
+export ANALYTICS_ID=${KOKORO_KEYSTORE_DIR}/72743_analytics_id
+export GITHUB_TOKEN=${KOKORO_KEYSTORE_DIR}/72743_paflynn_github_token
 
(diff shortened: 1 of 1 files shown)


kokoro/unix/release.sh
```diff
@@ -30,12 +30,9 @@
...
 
-## Assert required environment variables
-
-[ -z "$ANALYTICS_ID" ] && \
-    echo "ERROR: Releasing requires the ANALYTICS_ID environment variable" && exit 1;
-[ -z "$GITHUB_TOKEN" ] && \
-    echo "ERROR: Releasing requires the GITHUB_TOKEN environment variable" && exit 1;
+## Export required environment variables
+export ANALYTICS_ID=${KOKORO_KEYSTORE_DIR}/72743_analytics_id
+export GITHUB_TOKEN=${KOKORO_KEYSTORE_DIR}/72743_paflynn_github_token
```
chanseokoh: Double-checking: the variables hold the path for the files, not the content?

etanshaul: hmm. no I they hold the content. I may have confused how this works then

chanseokoh: OK, so I guess it should be like `MY_VAR=$(cat file.txt)` or `MY_VAR=$(< file.txt)`. (Not tested.)

etanshaul: Right. Now I’m wondering if it would have been better to leave the env var exporting and this cat’ing on the backend Kokoro scripts instead of here. Probably doesn’t matter much

chanseokoh: Kokoro logs the contents of the injected environment variables, so I think it's better to hide it. Moreover, it's an unnecessary detour (you already have the contents as files in the machine, why do you want to inject the contents again using a different method?), which also requires config on the backend, and I guess the simpler and shorter the config, the better.

etanshaul: @chanseokoh I updated the script to cat the file contents to the env var

PR #99, opened 2018-12-14
Update compiler.xml metadata

etanshaul:
@ivanporty have you noticed this as well? I'm not sure why this started happening (all of these modules are not submodules of `google-container-tools-intellij`)


ivanporty:
I don't see this changes on my IDE. Could it be minor differences between versions/Gradle/etc., and if so, do we need this at all in version control? I think if this file is missing and Gradle is used, nothing changes with the project?

etanshaul:
closing. its easier to just open a new pr for this

PR #111, opened 2018-12-18
Use parsed YAML structure to better detect valid Skaffold files.

### LINKED ISSUE #96: Use actual YAML parsing to better detect Skaffold files
Skaffold profiles support (#95) introduced YAML format basic parsing. It can be re-used to better identify valid Skaffold YAML file by parsing the content and detecting `apiVersion` instead of just checking the first line for regular expression.

ivanporty:
Fixes #96.

Instead of non-YAML regex, use YAML parser based detection for easier detection of valid Skaffold YAML files.

### DIFF
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFileService.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFileService.kt
@@ -44,6 +44,8 @@ class SkaffoldFileService {
                     val skaffoldYaml = SkaffoldYamlConfiguration(this)
                     return skaffoldYaml.apiVersion?.startsWith(SKAFFOLD_API_HEADER) == true
                 } catch (ex: Exception) {
+                    // We don't care about I/O or scan exceptions here since we only need to know if
+                    // the YAML file was in the proper format.
                     return false
                 }
             }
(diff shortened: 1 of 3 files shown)


skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldYamlConfiguration.kt
```diff
@@ -37,6 +37,9 @@ class SkaffoldYamlConfiguration(skaffoldYamlFile: VirtualFile) {
         )
     }
 
+    /** apiVersion of Skaffold configuration file, in the form of skaffold/v{number} */
+    val apiVersion: String? = skaffoldYamlMap["apiVersion"].toString()
```
etanshaul: what happens if `skaffoldYamlMap["apiVersion"]` doesn't exist?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldFileService.kt
```diff
@@ -47,12 +40,11 @@ class SkaffoldFileService {
...
             if (!isDirectory && fileType is YAMLFileType && isValid) {
-                val inputStream: InputStream = ByteArrayInputStream(contentsToByteArray())
-                inputStream.use {
-                    val scanner = Scanner(it)
-                    // consider this YAML file as Skaffold when first line contains proper API version
-                    return scanner.hasNextLine() &&
-                        SKAFFOLD_API_HEADER_PATTERN.matcher(scanner.nextLine()).find()
+                try {
+                    val skaffoldYaml = SkaffoldYamlConfiguration(this)
+                    return skaffoldYaml.apiVersion?.startsWith(SKAFFOLD_API_HEADER) == true
```
etanshaul: no need for the `== true`

etanshaul: ahh right. I guess the other way do deal with a `Boolean?` here would be to do `...startsWith(..) ?: false`. But what you have is clearer.

PR #119, opened 2019-01-07
Set up analytics panel and preference persistence

etanshaul:
more progress towards #47 

- Creates the Google menu item and the Usage Tracker menu item in the IDE settings\
- Persists the user preference

image

**Interop with Google Cloud Tools plugin**
The GCT plugin also defines these menu items. As long as the IDs of the configured panels match, the platform will ensure that _only a single instance of the menu item will be created_. The IDs are set as follows:
"Google" menu item -> "google.settings"
"Usage Tracking" menu item -> "google.usagetracker.settings"

These IDs must be set identically in both plugins. With this, when both plugins are installed, only one of the plugins will contribute the menu items (based on the loading order of the plugins). Since both plugins persist the user preference with the same key, the selection is effectively shared across the plugins.

TODO:
- Update the GCT plugin's usage tracker menu item to have the same ID (the "Google" menu item already has an ID): https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/2316
- Setup the usage tracking notification infrastructure (on first install of the plugin) - next PR


etanshaul:
@patflynn to look at the wording. I will update the wording on the GCT plugin to match the wording we finalize here.

core/src/main/kotlin/com/google/container/tools/core/util/UsageTrackerBundle.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.intellij.CommonBundle
+import org.jetbrains.annotations.PropertyKey
+import java.util.ResourceBundle
+
+private const val BUNDLE_NAME = "messages.UsageTrackerBundle"
+
+/**
+ * Message bundle manager for usage tracking related featured.
+ */
+object UsageTrackerBundle {
```
ivanporty: Could we possibly merge the two bundles (core and usage tracker) into one bundle, probably just `core`? It feels very fine grained and mildly confusing since the core has only one string and probably the usage tracker will stay in the core module anyway? I believe we usually chose to have a single bundle per module in our cloud tools plugin.

core/src/main/kotlin/com/google/container/tools/core/settings/UsageTrackerConfigurable.kt
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+
+        usageTrackingPanel.border = IdeaTitledBorder(
+            UsageTrackerBundle.message("usage.tracking.panel.title"),
+            0,
+            Insets(0, 0, 0, 0)
+        )
+
+        return panel {
```
ivanporty: Could we return `usageTrackingPanel` directly instead of additional panel wrapper?

etanshaul: I tried, but I couldn't get the titled panel thing to appear (where is says "Usage Tracking"). I suspect this is because the outer panel is the entire panel of the menu item (and no title is allowed)? Not 100% sure why.

core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerManagerService.kt
```diff
@@ -18,32 +18,42 @@ package com.google.container.tools.core.analytics
...
+import com.intellij.ide.util.PropertiesComponent
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.components.ServiceManager
 
 /**
  * Application service that manages the status of usage tracking in the plugin.
  */
-class UsageTrackerManagerService {
+class UsageTrackerManagerService(
+    private val trackingPreferenceProperty: PropertiesComponent = PropertiesComponent.getInstance()
```
ivanporty: It might be useful to add a short parameter style comment for this, i.e. how is this used in a few words.

core/src/test/kotlin/com/google/container/tools/core/settings/UsageTrackerConfigurableTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    private lateinit var usageTrackerManagerService: UsageTrackerManagerService
+
+    private lateinit var usageTrackerConfigurable: UsageTrackerConfigurable
+
+    @Before
+    fun setUp() {
+        usageTrackerConfigurable = UsageTrackerConfigurable()
+    }
+
+    @Test
```
ivanporty: since UI state is modified here, it makes sense to use `@UITest` as well to protect against possible racing

core/src/test/kotlin/com/google/container/tools/core/settings/UsageTrackerConfigurableTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    fun `isModified is true when stored preference does not match checkbox selection`() {
+        // stored value is false
+        every { usageTrackerManagerService.isTrackingOptedIn() } answers { false }
+        // preference is selected
+        usageTrackerConfigurable.usageTrackerCheckbox.isSelected = true
+
+        Truth.assertThat(usageTrackerConfigurable.isModified).isTrue()
+    }
+
+    @Test
```
ivanporty: ditto `@UITest`

core/src/test/kotlin/com/google/container/tools/core/settings/UsageTrackerConfigurableTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    fun `isModified is false when stored preference matches checkbox selection`() {
+        // stored value is true
+        every { usageTrackerManagerService.isTrackingOptedIn() } answers { true }
+        // preference is selected
+        usageTrackerConfigurable.usageTrackerCheckbox.isSelected = true
+
+        Truth.assertThat(usageTrackerConfigurable.isModified).isFalse()
+    }
+
+    @Test
```
ivanporty: ditto @UITest

core/src/test/kotlin/com/google/container/tools/core/settings/UsageTrackerConfigurableTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    fun `when preference checkbox is selected apply stores the tracking selection`() {
+        usageTrackerConfigurable.usageTrackerCheckbox.isSelected = true
+        usageTrackerConfigurable.apply()
+
+        verify { usageTrackerManagerService.setTrackingOptedIn(true) }
+    }
+
+    @Test
```
ivanporty: ditto `@UITest`

core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerManagerService.kt
etanshaul: done

etanshaul:
@ivanporty thanks for the review. I think I addressed your comments so far.

Still need to get a solution to [my comment](https://github.com/GoogleContainerTools/google-container-tools-intellij/pull/119#discussion_r246067553)

etanshaul:
I'll address [that issue ](https://github.com/GoogleContainerTools/google-container-tools-intellij/pull/119#issuecomment-452362034) in the next PR so that we can focus more clearly on it

PR #121, opened 2019-01-09
Fix for shared panel bug for users of old GCT plugin

etanshaul:
fixes https://github.com/GoogleContainerTools/google-container-tools-intellij/pull/119#issuecomment-452387513

Will prevent multiple instances of the tracking panel from showing up when users are using old versions of the GCT plugin (where the panel does not have an ID configured and therefore the platform can't handle de-duplication for us).

More explanation is in the comments on the code.

### DIFF
--- a/core/src/main/kotlin/com/google/container/tools/core/settings/UsageTrackerConfigurableProvider.kt
+++ b/core/src/main/kotlin/com/google/container/tools/core/settings/UsageTrackerConfigurableProvider.kt
@@ -38,12 +38,12 @@ class UsageTrackerConfigurableProvider : ConfigurableProvider() {
          * have an ID configured for this panel. Since we only want a single instance of this panel
          * to appear for all installed Google plugins, this checks to ensure that it was not already
          * registered by the GCT plugin. Once all users migrate to newer versions of the GCT plugin,
-         * we can remove this.
+         * we can remove this. For this to work, both provider classes have to have the same name.
          */
         val canCreateConfigurable: Boolean =
             Configurable.APPLICATION_CONFIGURABLE.extensionList.filter {
                 it?.providerClass != null &&
-                    it.providerClass.endsWith("UsageTrackerConfigurableProvider")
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/com/google/container/tools/core/settings/UsageTrackerConfigurableProvider.kt
```diff
@@ -32,6 +34,21 @@ class UsageTrackerConfigurableProvider : ConfigurableProvider() {
...
+    override fun canCreateConfigurable(): Boolean {
+        /**
+         * This check is a workaround for the fact that the older versions of the GCT plugin do not
+         * have an ID configured for this panel. Since we only want a single instance of this panel
+         * to appear for all installed Google plugins, this checks to ensure that it was not already
+         * registered by the GCT plugin. Once all users migrate to newer versions of the GCT plugin,
+         * we can remove this. For this to work, both provider classes have to have the same name.
+         */
+        val canCreateConfigurable: Boolean =
+            Configurable.APPLICATION_CONFIGURABLE.extensionList.filter {
```
ivanporty: To be clear, this will get a complete list of all configurables for the whole IDE and then you check the plugin is alone (meaning it's the only one that registered the provider). With Cloud Tools present, the count is 2, correct?

etanshaul: exactly

etanshaul: if you feel the javadoc isn't clear enough I can update?

ivanporty: No I think the javadoc is great, I was just making sure the code reads correct (I looked up `APPLICATION_CONFIGURABLE` quickly), I think this is perfectly clear.

PR #124, opened 2019-01-10
Usage enabled check needs to be called lazily

etanshaul:
`UsageTracker.create()` initializes the implementation to either the noop version or the real version depending on if it should be enabled (checked via the manager callback).

Previously this was initializing the implementation immediately and so it was set only once and the implementation didn't change depending on if the user opts in or out in the settings.

This updates it to recreate each time the `UsageTracker` is fetched (and hence each time a ping is sent).

### DIFF
--- a/core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerProvider.kt
+++ b/core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerProvider.kt
@@ -19,6 +19,7 @@ package com.google.container.tools.core.analytics
 import com.google.cloud.tools.ide.analytics.UsageTracker
 import com.google.cloud.tools.ide.analytics.UsageTrackerManager
 import com.google.cloud.tools.ide.analytics.UsageTrackerSettings
+import com.google.common.annotations.VisibleForTesting
 import com.google.container.tools.core.PluginInfo
 import com.intellij.openapi.application.ApplicationInfo
 import com.intellij.openapi.application.PermanentInstallationID
@@ -35,7 +36,8 @@ import com.intellij.openapi.components.ServiceManager
  */
 class UsageTrackerProvider {
 
-    private val usageTrackerSettings: UsageTrackerSettings
+    @VisibleForTesting
+    var usageTrackerSettings: UsageTrackerSettings
 
     init {
         val usageTrackerManagerService: UsageTrackerManagerService =
--- a/core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerProvider.kt
+++ b/core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerProvider.kt
@@ -17,6 +17,7 @@
 package com.google.container.tools.core.analytics
 
 import com.google.cloud.tools.ide.analytics.UsageTracker
+import com.google.cloud.tools.ide.analytics.UsageTrackerManager
 import com.google.cloud.tools.ide.analytics.UsageTrackerSettings
 import com.google.container.tools.core.PluginInfo
 import com.intellij.openapi.application.ApplicationInfo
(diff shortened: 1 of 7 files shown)


core/src/main/kotlin/com/google/container/tools/core/analytics/UsageTrackerProvider.kt
```diff
@@ -51,10 +52,16 @@ class UsageTrackerProvider {
...
-        usageTracker = UsageTracker.create(settings)
     }
 
+    /**
+     * Returns the appropriate implementation of [UsageTracker]. The implementation will change -
+     * from either a noop version, or the real version - depending on the value returned by the
+     * [UsageTrackerManager] callback which will depend on the current user selected tracking
+     * preference.
+     */
+    val usageTracker: UsageTracker get() = UsageTracker.create(usageTrackerSettings)
```
ivanporty: I don't recall directly from the previous PR - do we unit test this, i.e. make sure changed settings result in a valid usage tracker implementation returned?

etanshaul: I was looking at this. But the `Usage Tracker` implementations (both of which are in the analytics library) are not public, which makes it tough to unit test here. If you had any other approaches I'm open

etanshaul: I made them not public on purpose.

etanshaul: Actually, I could get the class name string of the impl and test that its the one we expect. Let me update

etanshaul: > Just to make it clear, we create settings once in init, but create new UsageTracker instance, and the way it will pick up the new settings is via the manager field in the settings, right? Once user changes the settings, UsageTracker picks up the new value in its constructor?

Correct. The UsageTracker getter will reinitialize the UsageTracker each time a ping is sent. This initialization (if you look at the `create` method) is based on the `UsageTrackerManager#isTrackingEnabled` callback which will now be called each time to ensure that the current user settings are taken into account

ivanporty: I would say this looks mildly confusing, as it's a bit hard to figure out why `create` method returns something different when we pass the same `settings` all the time, unless you dig deeper. How about passing `UsageTrackerManager#isTrackingEnabled` directly as a parameter into create - if I understand correctly, this is the only "mutable control bit"?

etanshaul: hmm. I guess if I did that, I could get rid of the `UsageTrackerManager` callback altogether as it wouldn't be needed. I could just pass in `isTrackingEnabled`..

etanshaul: something like this?
```
UsageTracker#create(UsageTrackerSettings settings, boolean isTrackingEnabled)
```

ivanporty: Just a bit confusing, in general it's easy to see what's happening behind the scenes and this code is not accessed / changed much anyway. I guess we may use a enum value instead of a boolean (like ENABLED, NO_OP) etc. It feels this PR is not a good place to do that, we may file a small enhancement on ide-analytics instead and come back to it later if we have more time for refactoring.

etanshaul: that sounds good. let me add some test coverage then we can merge and refactor separately.

etanshaul: added tests. ptal

PR #125, opened 2019-01-11
Plugin initialization component and telemetry

etanshaul:
Creates a plugin initialization component and initializes feature telemetry.


src/main/kotlin/com/google/container/tools/init/PluginInitComponent.kt
```diff
@@ -0,0 +1,46 @@
...
+        /* Initialize feature usage tracking. Only do this if:
+           1. The Google Cloud Tools plugin is not installed
+           2. Usage tracking is available
+           3. A tracking preference has not already been recorded
+         */
+        if (!isGctInstalled &&
+            UsageTrackerManagerService.instance.isUsageTrackingAvailable() &&
+            !UsageTrackerManagerService.instance.hasUserRecordedTrackingPreference()
+        ) {
+            UsageTrackerManagerService.instance.setTrackingOptedIn(true)
```
ivanporty: So this essentially sets tracking to be enabled by default if all conditions are met, as we discussed, right?

etanshaul: yes

src/main/kotlin/com/google/container/tools/init/PluginInitComponent.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.intellij.openapi.extensions.PluginId
+
+/**
+ * Performs initialization tasks when the plugin is loaded.
+ */
+class PluginInitComponent : BaseComponent {
+
+    override fun initComponent() {
+        val isGctInstalled =
+            PluginManager.isPluginInstalled(PluginId.getId(PluginInfo.GOOGLE_CLOUD_TOOLS_PLUGIN_ID))
```
ivanporty: Does the loading order guarantee we can check this condition regardless of which plugin gets loaded first?

etanshaul: I believe so. This check is that the plugin is "installed" not that its "loaded". So a plugin is considered installed even before load. All my testing confirms this.

PR #128, opened 2019-01-15
Change plugin id to more readable format

etanshaul:
This ID can be surfaced to the user. One example is when using `externalDependencies.xml` file declaring this plugin as a dependency; in this case the IDE will show a notification showing this ID suggesting to install the plugin.


etanshaul:
@patflynn fyi - this is relevant to a recent thread about `externalDependencies.xml` and this ID being shown to the user.

etanshaul:
right. if you look at various plugins in the repository you can see different strategies used. From java package style to came case etc.

Issue #52, opened 2018-10-25
Support settings for Skaffold default image repo


ivanporty:
Skaffold now supports concept of *default image repository*, the one that is used for all Skaffold deployments regardless of what is configured in skaffold configuration for build stage. This allows users to try out other developer pipelines without having to worry about changing build step and image/repo name.

We can add this to individual run configuration or as a global setting.

See https://github.com/GoogleContainerTools/skaffold/pull/1057 for more details.

etanshaul:
+1. was just testing out a non-trivial project and this would be useful. I'm in favor of adding it to the individual run configs

PR #130, opened 2019-01-23
UI and settings for image repository override.

ivanporty:
Work towards #52.

See #52 and GoogleContainerTools/skaffold#1057 for more details.
This PR sets up UI and settings along with unit tests for image repository override support.
The UI:

image-repo-ui

### DIFF
--- a/skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/ui/BaseSkaffoldSettingsEditor.kt
+++ b/skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/ui/BaseSkaffoldSettingsEditor.kt
@@ -24,7 +24,9 @@ import com.intellij.openapi.options.ConfigurationException
 import com.intellij.openapi.options.SettingsEditor
 import com.intellij.openapi.vfs.LocalFileSystem
 import com.intellij.openapi.vfs.VirtualFile
+import com.intellij.ui.TitledSeparator
 import com.intellij.ui.border.IdeaTitledBorder
+import com.intellij.ui.components.JBTextField
 import com.intellij.ui.layout.panel
 import com.intellij.util.ui.UIUtil
 import java.awt.Insets
@@ -52,6 +54,8 @@ open class BaseSkaffoldSettingsEditor<T : AbstractSkaffoldRunConfiguration>(
     @VisibleForTesting
     val skaffoldProfilesComboBox = SkaffoldProfilesComboBox()
 
+    private val overrideImageRepoTextField = JBTextField()
+
     protected lateinit var basePanel: JPanel
 
     private val extensionComponents: MutableMap<String, JComponent> = mutableMapOf()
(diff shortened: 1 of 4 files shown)


etanshaul:
nice! Suggestions for the wording:

1) "override image repository" -> image repository prefix (or default image repository prefix?)
2) placeholder text -> still thinking about this one.. it wasn't clear to me looking at it which part I needed to enter. perhaps the update to 1) will help

etanshaul:
May be long, but how about this placeholder text:

`e.g. docker.io or gcr.io. This will prefix all image names in the k8s manifests with this repository name.`

ivanporty:
Thanks for the ideas.
1. I think this is treated as a prefix, but it should be a valid repository "identifier" in order for this to work. How about `Custom image repository:`?

2. For longer text, how about using a tooltip with a helper text you suggested? Again, just adding "docker.io" or "gcr.io" won't make it work unless proper Docker Hub username or GCP project name is set, hence the prompt I tried to improvise. Open to more suggestions on this.

ivanporty:
For (2), we might try something similar to `docker.io/{account}, gcr.io/{project-id}, etc.` instead of just two options with `or` which might mislead some people to think only these two would actually work.

ivanporty:
Agree, this looks concise. Let's try that, I'll do the changes and attach an updated screenshot.

ivanporty:
screen shot 2019-01-24 at 4 05 03 pm

ivanporty:
@etanshaul PTAL, looks pretty clear I think!

etanshaul:
nice!

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/ui/BaseSkaffoldSettingsEditor.kt
```diff
@@ -77,10 +82,18 @@ open class BaseSkaffoldSettingsEditor<T : AbstractSkaffoldRunConfiguration>(
...
+
+            row {
+                TitledSeparator(message("skaffold.image.options.subtitle"))(grow)
+            }
+            row(message("skaffold.override.image.repo")) { overrideImageRepoTextField(grow) }
         }
 
         basePanel.border = IdeaTitledBorder(editorTitle, 0, Insets(0, 0, 0, 0))
 
+        overrideImageRepoTextField.emptyText.text = "e.g. docker.io, gcr.io"
```
etanshaul: message bundle?

patflynn:
so if somebody uses this default image repo and omits the full url in their manifests, doesn't that mean that skaffold will fail from the CLI?

patflynn:
should have read the full thread, so this is a skaffold feature. deleting comment.

etanshaul:
it wont fail if they use the `--default-image-repo flag` (which is what this feature is doing).

patflynn:
I see. Can that only be a flag or can it be configured in the yaml?

etanshaul:
pretty sure its only as a flag, currently

patflynn:
yeah using the flag makes sense. 👍

PR #136, opened 2019-01-29
Spring boot getting started example for container tools

### LINKED ISSUE #133: Add Java SpringBoot quickstart example into the project repo
We should probably have our own "Hello World" style example for SpringBoot RESTful application, and use it as a reference for quick start with the plugin and its features in the tutorial / getting started documentation, rather than creating everything from scratch or using external getting started project.

It should definitely use Jib and can be based on https://github.com/GoogleContainerTools/skaffold/tree/master/examples/jib with a few enhancements.

ivanporty:
Fixes #133.

Initial Spring Boot Hello World example to use with basic Container Tools Skaffold features and for quick Github based documentation. As we continue with the quick start documentation, we might modify and expand this example a bit more to showcase more plugin features.

### DIFF
--- a/examples/hello-spring-boot/.gitignore
+++ b/examples/hello-spring-boot/.gitignore
@@ -4,5 +4,5 @@ target
 .settings
 .vscode
 .idea/*
-!.idea/
 !.idea/externalDependencies.xml
+


etanshaul:
awesome!

I think we should consider adding these two things too:
- A dependency on this plugin so that when the user clones the repo without the plugin installed, then it prompts them to install it (see https://github.com/GoogleCloudPlatform/google-cloud-intellij/blob/master/.idea/externalDependencies.xml).
- (I could be swayed either way on this one), the .idea metadata with the Skaffold run configurations already configured? May run into issues with this if we ever change the run configurations and the user is running a version of the plugin incompatible with the source of this example

ivanporty:
Great suggestion, I'll add dependency on the plugin and test on IDE without it.

As for the metadata, I thought it would be easier for a user to open this project from scratch, and even in the other IDE to see the difference. Also, as part of the `Getting Started` section, I want to show a screenshot of Skaffold prompt popping up, and how it enables one-click run targets. WDYT?

etanshaul:
> As for the metadata, I thought it would be easier for a user to open this project from scratch, and even in the other IDE to see the difference. Also, as part of the Getting Started section, I want to show a screenshot of Skaffold prompt popping up, and how it enables one-click run targets. WDYT?

sgtm. I'm fine leaving out the run config metadata.

regarding testing out the plugin dependency - I'm not sure how this will look in your testing given that we don't yet the plugin published to the JB repo. Even if it doesn't work we can still merge it though now.

ivanporty:
It works, just shows empty list of plugins when you click `Install plugins...` but I think this is perfect to be sure example is opened in properly setup IntelliJ/ other IDE:

plugin-required-example

examples/hello-spring-boot/.gitignore
```diff
@@ -0,0 +1,8 @@
+target
+.classpath
+.project
+.settings
+.vscode
+.idea/*
+!.idea/
```
etanshaul: Do you need ‘!.idea’ given that your are excluding the xml file itself?

ivanporty: let me check, I was wrestling with these rules a bit and might have over-complicated.

ivanporty: You're correct, not necessary! Removed it, thanks!

ivanporty:
Thanks for the review, definitely good idea to have a simple README here.

PR #141, opened 2019-02-01
Automates creating prerelease distributions as part of the CI

### LINKED ISSUE #139: Make pre-release plugin build available for early access
(Originally suggested by @etanshaul)

Currently there is no easy way to install the preliminary version of the plugin for testing and early access before we have it available in Jetbrains repository. We can use Kokoro to automate plugin build (`buildPlugin` build task) and copying the resulting zip distributive into a directory (at the root of the project in a `prerelease` folder for example) on every commit to master. This would avoid the initial build step.

etanshaul:
fixes #139 

To be merged only after backend Kokoro config updated


kokoro/unix/continuous.sh
```diff
@@ -0,0 +1,38 @@
...
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+cd github/google-container-tools-intellij
+
+# build the distribution and copy to the pre-release folder
+mkdir -p prerelease
+./gradlew buildPlugin
+cp build/distributions/*.zip prerelease
```
ivanporty: Will this commit the resulting archive into `prerelease` folder in GitHub too?

etanshaul: no. I realized this too. I need to fix

etanshaul: makes me think that this would be better as part of the build process. so that it gets included as a single commit with the actual code changes

etanshaul: or possibly as part of the GitHub release mechanism. looking into this

etanshaul:
closing for now. as I don't think this is a great approach. the best option is to just use the JB channels once we can

PR #158, opened 2019-02-12
Skaffold system check

ZariaHoward:
Checks to see if Skaffold is on system when the user tries to Deploy.


etanshaul:
Some additional tips:

**Meaningful commit message** 
The message should be descriptive of the problem (and potentially the fix as well). See previous commits in the repo for some examples. In this case, a possible message could be `Checks to ensure Skaffold is on the PATH and shows error message if not`

**PR Description** 
Start the PR with `fixes #[bug number here]` on the first line, if there is an applicable issue. GitHub will then close the associated issue automatically when merging into master. 

Add sufficient detail to the PR description to describe your solution. Also, screenshots are always helpful when there are any UI facing changes. In this case, a screenshot of the log message would be helpful.

**Build Checks**
As @ivanporty mentioned, you can run `gradlew check` locally to ensure that all of the checks pass, including unit tests. If they are not passing locally then they will likely not pass our Kokoro CI.

**Unit Tests**
Add unit tests whenever possible. In some cases it may not make sense, but the reasons should be discussed / called out. See some of our existing unit tests in this repo for helpful patterns (for example, we use the `ContainerToolsRule` test rule for our unit tests).

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -101,4 +106,18 @@ abstract class AbstractSkaffoldRunConfiguration(
 
         XmlSerializer.serializeInto(this, element)
     }
+
+    @Throws(RuntimeConfigurationException::class)
```
etanshaul: Probably don't want to commit these changes?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -85,6 +85,8 @@ abstract class SkaffoldExecutorService {
         }
 
         try {
+            //this is where that whole no skaffold directory thing is happening
+            // Wait why isnt going into the catch
```
etanshaul: oh it seems like this may not be ready yet for review? if so, its fine to put out the PR, just don't assign it for review until it is ready.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
```diff
@@ -60,6 +64,24 @@ class SkaffoldCommandLineState(
             throw ExecutionException(message("skaffold.no.file.selected.error"))
         }
 
+        //call skaffold version --- look at createProcess function, the command list lets you know if skaffold isn't on the system by process of elimination
+        if (Paths.get("skaffold").exists()) {
```
etanshaul: I think a better way to check that Skaffold is on the PATH is to check the system path directly and see if `skaffold` is there. Its generally advised not to use exception control flow in this way. Something like this (with comments added for clarity):

```
fun isExecutableInPath(executable: String): Boolean {
   val path = System.getenv("PATH") // get the full path string
   return path.split(File.pathSeparator) // split the path string by the path separator
      .asSequence() // use sequence for lazy evaluation
      .map { it + File.separator + executable } // convert each to a possible path to the executable
      .any {
         File(it).exists() // does the file exist?
      }
}
```

then calling `isExecutableInPath("skaffold")` should do the check

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
```diff
@@ -60,6 +64,24 @@ class SkaffoldCommandLineState(
...
+            val testSkaffoldCommand = Paths.get("skaffold").toString() + " " + (executionMode.modeFlag)
```
etanshaul: one reason the build is failing is because this line is over 100 chars. you can see this by running `./gradlew check`. these currently have to be fixed manually because `./gradlew spotlessApply` does not yet fix line length violations

etanshaul: by "manually" I mean you can run a reformat in intellij: code -> reformat code

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -21,14 +21,19 @@ import com.google.container.tools.skaffold.run.ui.SkaffoldDevSettingsEditor
 import com.google.container.tools.skaffold.run.ui.SkaffoldSingleRunSettingsEditor
 import com.intellij.execution.Executor
 import com.intellij.execution.configurations.ConfigurationFactory
+import com.intellij.execution.configurations.ConfigurationPerRunnerSettings
```
etanshaul: Some of these imports are unused also causing the build to fail. an easy way to fix this is to run "optimize imports":

`Code -> Optimize Imports`

etanshaul: `./gradlew spotlessApply` should also fix that _I think_

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurationsTest
```diff
@@ -0,0 +1,167 @@
...
+import io.mockk.mockkStatic
+import io.mockk.slot
+import org.jdom.Element
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import java.io.File
+
+/** Unit tests for [SkaffoldCommandLineState] */
+class SkaffoldCommandLineStateTest {
```
etanshaul: the class name should typically match the file name. did you mean to call this class `SkaffoldRunConfigurationsTest`?

etanshaul: also it looks like you added it as a text file instead of a `.kt` file. 

When creating a new Kotlin file in the IDE, use:
`right click on the folder` -> new -> Kotline File/class

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -43,7 +43,7 @@ abstract class SkaffoldExecutorService {
     }
 
     /** Path for Skaffold executable, any form supported by [ProcessBuilder] */
-    protected abstract var skaffoldExecutablePath: Path
+    open var skaffoldExecutablePath: Path = Paths.get("skaffold")
```
etanshaul: this change shouldn't be necessary. the `DefaultSkaffoldExecutorService` is overriding this and setting the value

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
```diff
@@ -60,6 +64,24 @@ class SkaffoldCommandLineState(
...
+            try {
+                SkaffoldProcess(
+                    SkaffoldExecutorService.instance.createProcess(
+                        File(projectBaseDir.path),
+                        listOf(testSkaffoldCommand)
+                    ),
+                    commandLine = testSkaffoldCommand//skaffold executable path
+                )
+            }catch (e: Exception){
+                throw ExecutionException("skaffold path in system is incorrect")
```
etanshaul: any time we have UI visible messages in the code, we should put them in the message bundle. see line 64 for an existing example of this:

`throw ExecutionException(message("skaffold.no.file.selected.error"))`

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -106,18 +111,33 @@ abstract class AbstractSkaffoldRunConfiguration(
...
-        throw RuntimeConfigurationException("Hitting function2222")
     }
 
 
-    override fun checkSettingsBeforeRun(){
-        throw RuntimeConfigurationException("Hitting function")
+    companion object {
+        var skaffoldPath: Path? = Paths.get("skaffold")
+        @JvmStatic
+        fun checkConfiguration(project:Project) {
```
etanshaul: can this just live in the class itself instead of putting it in a companion object?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -106,18 +111,33 @@ abstract class AbstractSkaffoldRunConfiguration(
...
+            val projectBaseDir: VirtualFile? = project.guessProjectDir()
+            if (skaffoldPath?.exists() == true && projectBaseDir != null ) {
+                val testSkaffoldCommand = skaffoldPath.toString() + " " + (SkaffoldExecutorSettings.ExecutionMode.SINGLE_RUN.modeFlag)
```
etanshaul: don't forget to update this to either do a version check. or update the path checking strategy using the code snippet suggestion I gave in another comment

etanshaul:
@ivanporty for review as well

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
etanshaul: one option to help with testing:

move the path check to `SkaffoldFileService`. This is an applicationService defined in `skaffold.xml`. You can swap these out in unit tests with a mock using the `@TestService` annotation. I can walk you through this.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
ZariaHoward: Apparently Kotlin doesn't have static members of abstract classes from what I read? I wanted the function to be able to be called without creating an instance, an according to the docs that is where companion objects come in.

etanshaul: makes sense.

see my other suggestion about moving this utility to `SkaffoldFileService`. This should help better structure things (since you are using this check in several contexts); as well as help make unit testing easier.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
etanshaul: sorry - I think `SkaffoldExecutorService` is the better place for this

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
```diff
@@ -101,4 +111,33 @@ abstract class AbstractSkaffoldRunConfiguration(
...
         XmlSerializer.serializeInto(this, element)
     }
+    override fun checkConfiguration(){
+        checkConfiguration(project)
+
+    }
+
+
+    companion object {
+        var skaffoldPath: Path? = Paths.get("skaffold")
```
ivanporty: We probably don't need this at all - at the end we call `SkaffoldExecutorService.instance.createProcess` and if it works, it should work for general case as well.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurations.kt
ivanporty: +1 on relocating this as a normal member function to `SkaffoldExecutorService`

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
```diff
@@ -60,6 +62,12 @@ class SkaffoldCommandLineState(
             throw ExecutionException(message("skaffold.no.file.selected.error"))
         }
 
+        try {
+            AbstractSkaffoldRunConfiguration.checkConfiguration(environment.project)
```
ivanporty: So if we put this into `SkaffoldExecutorService`, we have the following `SkaffoldExecutorService.instance.checkConfiguration`, with the added benefit of easier mocking and dynamic service lookup.

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldRunConfigurationsTest.kt
```diff
@@ -0,0 +1,116 @@
...
+    fun `A runtime warning is not shown if skaffold is in the system PATH`() {
+
+        val testObj = mockk<AbstractSkaffoldRunConfiguration.Companion>()
+
+        every { testObj.skaffoldPath } returns Paths.get("skaffold")
+
+        testObj.checkConfiguration(mockExecutionEnvironment.project)
+
+    }
+}
```
ivanporty: I think we add a newline at the end of our files - I believe this is a common GitHub practice.

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/ui/BaseSkaffoldSettingsEditor.kt
```diff
@@ -113,6 +114,10 @@ open class BaseSkaffoldSettingsEditor<T : AbstractSkaffoldRunConfiguration>(
             throw ConfigurationException(message("skaffold.invalid.file.error"))
         }
 
+        if(true){
```
ivanporty: This is probably a test code, to be removed once the check is in place?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
etanshaul: thanks @ivanporty! @ZariaHoward and I discussed doing exactly this

skaffold/src/main/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineState.kt
etanshaul: @ivanporty had a great point that this doesn't check that the file is executable. I would suggest ammending the last line:

`File(it).exists() && File(it).canExecute()`

skaffold/src/main/resources/messages/SkaffoldBundle.properties
```diff
@@ -42,3 +42,4 @@ skaffold.override.image.repo.tooltip=<html>The default image repository allows y
 skaffold.no.file.selected.error=Skaffold configuration file is not selected.
 skaffold.invalid.file.error=Skaffold configuration file doesn't exist or invalid.
 skaffold.corrupted.run.settings=Your Skaffold run configuration is corrupted. Please re-create it to fix.
+skaffold.not.on.system.error=Skaffold path in system is incorrect. Install Skaffold. See required dependencies <a href="https://github.com/GoogleContainerTools/google-container-tools-intellij#prerequisites-and-required-dependencies">here</a>:
```
etanshaul: nit suggestion:

`Cloud not detect Skaffold installation. Please install Skaffold. See...`

skaffold/src/test/kotlin/com/google/container/tools/skaffold/DefaultSkaffoldExecutorServiceTest.kt
```diff
@@ -226,4 +227,16 @@ class DefaultSkaffoldExecutorServiceTest {
 
         assertThat(result.commandLine).isEqualTo("skaffold dev --filename test.yaml")
     }
+
+    @Test
+    fun `isSkaffoldAvailable returns true when skaffold is available`() {
+        assertThat(defaultSkaffoldExecutorService.isSkaffoldAvailable()).isTrue()
```
etanshaul: this will only work if Skaffold is available on the system running this unit test (will probably fail in CI since skaffold is not installed on those machines). you probably want to mock the path as you do in the test below. You'll also then need to mock a file so that `isSkaffoldAvailable` finds the skaffold executable.

for this you can use the `@TestFile` annotation:

at the class level add the following (which creates a test file named `skaffold` that gets cleaned up automatically when the test completes):
```
@TestFile(name = "skaffold")
private lateinit var testSkaffoldFile: File
```

Then you can mock the system path to return `testSkaffoldFile.parent`. This simulates having the test file's directory on the system path.

skaffold/src/test/kotlin/com/google/container/tools/skaffold/run/SkaffoldCommandLineStateTest.kt
```diff
@@ -141,4 +143,32 @@ class SkaffoldCommandLineStateTest {
...
+            SkaffoldExecutorSettings.ExecutionMode.DEV
+        )
+
+        expectThrows(
+            ExecutionException::class,
+            ThrowableRunnable { skaffoldCommandLineState.startProcess() })
+    }
+
+    @Test
+    fun `A run error isn't thrown if skaffold is in the system PATH`() {
```
etanshaul: I’d remove all the single quotes from these test names. It was causing weird behavior when running the tests from the IDEAs JUnit runner.

ZariaHoward:
screenshot from 2019-02-14 15-36-18

skaffold/src/main/resources/messages/SkaffoldBundle.properties
```diff
@@ -42,3 +42,4 @@ skaffold.override.image.repo.tooltip=<html>The default image repository allows y
 skaffold.no.file.selected.error=Skaffold configuration file is not selected.
 skaffold.invalid.file.error=Skaffold configuration file doesn't exist or invalid.
 skaffold.corrupted.run.settings=Your Skaffold run configuration is corrupted. Please re-create it to fix.
+skaffold.not.on.system.error=Could not detect Skaffold installation. Please install Skaffold. See required dependencies <a href="https://github.com/GoogleContainerTools/google-container-tools-intellij#prerequisites-and-required-dependencies">here</a>:
```
ivanporty: Does this link work in the message? I see it ends with a colon, is this intentional?

skaffold/src/test/kotlin/com/google/container/tools/skaffold/DefaultSkaffoldExecutorServiceTest.kt
```diff
@@ -226,4 +231,18 @@ class DefaultSkaffoldExecutorServiceTest {
...
+
+    @Test
+    fun `isSkaffoldAvailable returns true when skaffold is available`() {
+        mockkStatic(System::class)
+        every { System.getenv("PATH") } answers { testSkaffoldFile.parent }
+        assertThat(defaultSkaffoldExecutorService.isSkaffoldAvailable()).isTrue()
+    }
+
+    @Test
+    fun `isSkaffoldAvailable returns false when skaffold is not available`() {
```
ivanporty: It might also be a good idea to add a similar unit test, but with actual list of directories in mocked PATH, not only empty PATH, so we know for sure the parsing works.

etanshaul:
@ZariaHoward turns out the windows test failures are due to the `File#setExecutable`..as tested on my windows vm. I'm not exactly sure why yet.

etanshaul:
actually, it might be the `mockkStatic`

etanshaul:
alternative to using mockkStatic:

in `SkaffoldExecutorService#isSkaffoldAvailable`, extract `System.getenv("PATH")` to a separate function so that it can be mocked. I.e.:

```
fun getSystemPath(): String = System.getenv("PATH")
```

Then in the test, remove all `mockkStatic`, and instead of mocking System.getenv, mock:

`every { defaultSkaffoldExecutorService.getSystemPath() } answers { [whatever you want] }`

skaffold/src/main/resources/messages/SkaffoldBundle.properties
etanshaul: It probably should not end with a colon. Also the link does not work. Probably need to attach a URL handler.

etanshaul: I'd suggest, if its not easy to figure out how to get the URL to work here, to just change it from a hyperlink to just print the url (so the user can copy/paste it), and open a separate issue to track getting the link working.

ivanporty: Agree, might be not easy here. We can also think of some sort of notification - they are more flexible.

ZariaHoward:
screenshot from 2019-02-21 15-18-26

etanshaul:
looks good. commit the code updates?

common-test-lib/src/main/kotlin/com/google/container/tools/test/ContainerToolsRule.kt
```diff
@@ -133,6 +133,7 @@ class ContainerToolsRule(private val testInstance: Any) : TestRule {
                 }
             }
 
+            file.setExecutable(true)
```
ivanporty: It does not seem necessary for each test file we are creating, could we add a boolean parameter (optional, defaults to false) to mark file as executable? Or some other way to mark file executable in the unit test itself?

skaffold/src/main/kotlin/com/google/container/tools/skaffold/SkaffoldExecutorService.kt
```diff
@@ -42,9 +48,41 @@ abstract class SkaffoldExecutorService {
...
+            type,
+            NotificationListener.UrlOpeningListener(true)
+        )
+
+
+
     /** Path for Skaffold executable, any form supported by [ProcessBuilder] */
     protected abstract var skaffoldExecutablePath: Path
 
+    open fun getSystemPath(): String = System.getenv("PATH")
```
etanshaul: no need to put this in the interface. you can keep only in the Default impl

